PR #23, opened 2023-03-23
Document backend authentication mechanisms

KrisJordan:
Can you review this documentation and let me know where additional clarity should be added?

### DIFF
--- a/backend/api/authentication.py
+++ b/backend/api/authentication.py
@@ -1,3 +1,60 @@
+"""User authentication over HTTP via Headers and/or HTTP JWT Bearer Tokens.
+
+This module provides a `registered_user` dependency injection function for other routes
+to use to both ensure a user is authenticated and resolve to the logged in User's model.
+Further, this module provides the routes and logic for backend authentication.
+
+The implementation of auth routes are nuanced due to UNC Cloud Apps' Single Sign-On (SSO)
+proxy service. In production, there is a proxy sitting in front of the app that
+integrates with UNC's SSO/Shibboleth service for authentication. For more information
+on this proxy service, see the [official CloudApps documentation](https://help.unc.edu/sp?id=kb_article_view&sysparm_article=KB0011256)
+
+In production, the proxy intercepts all routes prefixed with `/auth`. Two paths follow:
+
+1. If the user is not logged in, the proxy redirects the user to the authentication server.
+2. If the user is logged in, the proxy sets the `uid` and `pid` headers to the user's
+    Onyen and PID, respectively, and forwards the request to our app.
+
+Once the request is forwarded to our app server, the `uid` and `pid` headers are used to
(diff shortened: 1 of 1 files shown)


backend/api/authentication.py
```diff
@@ -1,3 +1,60 @@
...
+
+In development, the proxy is not present. Instead, there are two options for authentication:
+
+1. If an unauthenticated user visits /auth in development, or staging, they are redirected
+   to the production `csxl.unc.edu/auth` route with an additional query parameter `origin`.
+    A. The production server authentication works as usual, but if the `origin` parameter is
+       detected alongside the SSO headers, the user will be redirected back to the `origin` 
+       server with a JWT `token` query parameter. This token is signed by the production server.
+    B. Back on the development/staging server, the `token` query parameter a backchannel request
+       is made to the production server to verify the token's validity. If the token is valid,
```
theyanaga, line 32: This sentence is a little unclear, maybe there's a typo or I am just reading it wrong. 

Is this supposed to be "Back on the development/staging server, a backchannel request is made to verify the token's validity?

backend/api/authentication.py
```diff
@@ -1,3 +1,60 @@
...
+       the development/staging server then issues a new `token` to the client that is signed
+       by the development/staging server. This token is then used for all subsequent requests.
+2. If an unauthenticated user visits /auth/as/{uid}/{pid} in development, they are authenticated
+    as the user with the given `uid` and `pid`, which are their ONYEN and PID, respectively. 
+    This route is only available in development mode.
+
+The router is mounted at `/auth` and provides the following endpoints:
```
theyanaga, line 39: Maybe moving the routes up to the top of the documentation would be a good idea? 

Students see the routes first with a high level overview, and then you dig deeper into each route, saying what it does.

theyanaga (approved):
Looks good! Just had one sentence that was rather unclear to me and then one suggestion to the feedback structure (optional)

backend/api/authentication.py
KrisJordan: Agreed, rewrote in the updated push:

    Back on the development/staging server, we need to verify that the token given to the route
    was actually signed by the production server. If we did not do this, a malicious user could
    simply generate a token and pass it to the development server to gain access. Thus, an HTTP
    request from the development/stage server is made to the production server's `/auth/verify` route
    to verify the token's validity.

backend/api/authentication.py
KrisJordan: Agreed, rewrote on the latest push: 

Back on the development/staging server, we need to verify that the token given to the route was actually signed by the production server. If we did not do this, a malicious user could simply generate a token and pass it to the development server to gain access. Thus, an HTTP request from the development/stage server is made to the production server's `/auth/verify` route to verify the token's validity

PR #24, opened 2023-03-24
Refactor backend user service to be idiomatic

KrisJordan:
This PR aims to do two things:

1. Add documentation to the profile editing routes
2. Refactor the user service to be more idiomatic

The latter is toward the end of this service being used as an example for developers to use as reference. Whereas the profile service endpoints expect the authenticated user to be the same as the user being created/updated in the user service, this service is more useful if we make it work more generally and idiomatically. The _subject_ is the user attempting to carry out the action and the _user_ is the resource the action applies to. This is the more common case developers will experience in features: _subject_ and _resource_ as two separate concerns. When profile editing, it's a special edge case where they are the same.

### DIFF
--- a/backend/api/profile.py
+++ b/backend/api/profile.py
@@ -45,7 +45,7 @@ def update_profile(
     than registered_user.
 
     ProfileForm is used here, rather than User, for similar registration-specific
-    purposes. Importantly, it does not include the ID.
+    purposes. Importantly, ProfileForm doesn't contain an ID field.
     """
     pid, onyen = pid_onyen
     user = user_svc.get(pid)
--- a/backend/api/profile.py
+++ b/backend/api/profile.py
@@ -1,15 +1,29 @@
+"""Profile API
+
+This API is used to retrieve and update a user's profile."""
+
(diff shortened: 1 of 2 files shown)


backend/api/profile.py
```diff
@@ -24,6 +38,15 @@ def update_profile(
...
     user_svc: UserService = Depends()
 ):
+    """Update a user's profile. If the user does not exist, create a new user.
+
+    Since the user is authenticated, we can trust the pid and onyen. However,
+    since the user may not be registered, we depend on authenticated_pid rather
+    than registered_user.
+
+    ProfileForm is used here, rather than User, for similar registration-specific
+    purposes. Importantly, it does not include the ID.
```
theyanaga, line 48: This perhaps could be phrased a bit better. 

I got it, but I can see some confusion from the wording "the ID." 

Maybe something along the lines of: "Importantly, ProfileForm doesn't contain an ID field"?

theyanaga (commented):
Looks good! Just one wording thing on the word ID, but otherwise I thought everything was pretty straight forward!

Nice stuff!

Issue #16, opened 2023-03-21
Bring documentation up to standards.


KrisJordan:
In the sprint toward being ready to release the project base to students on time, code that does not meet standards was shipped. This needs fix with hi-pri.

PR #25, opened 2023-03-24
Refactor backend user service to be idiomatic and documented (#24)

KrisJordan:
Factored out separate create vs. update methods.

Made the signature of create and update in the user service idiomatic (subject first, user model second) and improved permission check to have both feature-specific rule (a user can edit their own profile), as well as administrative permission check.


KrisJordan:
Merging into main for emergency purposes of being ready for class. This merge does not resolve issue #16, but makes progress.

PR #29, opened 2023-03-27
Fix redirect bug in production authentication

KrisJordan:
When performing delegated authentication, the origin host is passed
as a query parameter. This is then used to redirect the user back to
the correct page after authentication. However, in production, the
origin host was not being passed correctly, and the user was being
redirected back to the production hostname.


theyanaga (approved):
LGTM

Issue #49, opened 2023-09-01
Refactor "Org" to "Organization"


jadekeegan:
# Refactor "Org" to "Organization"
Update all instances of the abbreviation "org" to be "organization" instead.

## Why?
This is important to ensuring there is clarity throughout the code.

## Acceptance Criteria
- [x] Rename variables with "org" to be "organization".
- [x] Rename file names with "org" to be "organization".
- [x] Rename components and routes with "org" to be "organization".

PR #54, opened 2023-09-02
Refactor Organization Naming

jadekeegan:
This PR updates all instances of the abbreviation "org" to instead be written out fully as "organization". It also updates some erroneous usages of "organizations" instead of "organization", namely in the API routes/paths.

## Major Changes
- Updated every variable/component including "org" and some paths including "organizations" to say "organization".
- Renamed each file using the "org" abbreviation to say "organization".

## Testing

> *This PR resolves issue #49*


backend/api/organization.py
```diff
@@ -12,7 +12,7 @@
 __copyright__ = 'Copyright 2023'
 __license__ = 'MIT'
 
-api = APIRouter(prefix="/api/organizations")
+api = APIRouter(prefix="/api/organization")
```
ajaygandecha, line 15: This actually brings up a good point that @KrisJordan mentioned Friday, and something we should clarify. Since we are returning all organizations in this API route, it may be a good idea to keep this API route as `/api/organizations` to imply the expectation that a list should return from this.

ajaygandecha (changes requested):
@jadekeegan this looks great, very thorough pass! I do not see any remaining lingering usages of `org`. Only left one comment regarding API route naming conventions which may be good for us to review. Other than that, nice job!

backend/api/organization.py
ajaygandecha: (Also just a note here so we do not forget, if we do end up with `/api/organizations`, a few API calls in the frontend service may have to be reverted.)

ajaygandecha (approved):
LGTM! Great work!

frontend/src/app/admin/organization/organization-admin.service.ts
```diff
@@ -11,22 +11,22 @@ export class OrganizationAdminService {
...
     }
 
     /** Creates an organization
      * @param newOrg: Organization object that you want to add to the database
      * @returns {Observable<Organization>}
      */
-    createOrganization = (newOrg: Organization) => {
-        return this.http.post<Organization>("/api/organization", newOrg)
+    createOrganization = (newOrganization: Organization) => {
+        return this.http.post<Organization>("/api/organization", newOrganization)
```
atoneyd, line 22: Shouldn't this say api/organization**s** (adding an s at the end)

atoneyd (changes requested):

atoneyd (approved):
LGTM

briannata (approved):
Also LGTM!

Issue #56, opened 2023-09-06
Fix hover tooltip issues


atoneyd:
The hover tooltip is not working for certain descriptions that are truncated.

PR #57, opened 2023-09-06
Fix issue with tooltip not showing

atoneyd:
Fixed the following bug: On the organizations page, sometimes the tooltip would not show because of the way it was hard-coded to the length of characters in the description.

## Major Changes
- Added a function that checks whether or not the description height has been clipped, which is a more accurate indication of whether or not the tooltip should be shown
- Removed some unnecessary CSS

## Testing
Ran `honcho start`, made sure the tooltip showed up for all descriptions with ellipses.

> *This PR resolves issue #56 *


briannata (approved):
LGTM! All organization tooltips worked for me as expected when I ran `honcho start`.

Good work, Audrey!

PR #59, opened 2023-09-07
Organization API Paths Slugs (instead of IDs)

### LINKED ISSUE #53: Replace IDs with Slugs in Paths
# Replace IDs with Slugs in Paths
Replace the ID key in paths/routes with the organization slug.

## Why?
This is important to improving the overall appearance of the live site by routing to organization details with slugs rather than their numerical keys. It also helps simplify getting path parameters as everything will already be in strings.

## Acceptance Criteria
- [x] Add slugs to each organization in the development data.
- [x] Add an API route to get organization by slug.
- [x] Remove the API route to get organization by name.
- [x] Update the frontend to use slugs in the organization route.

briannata:
This pull request refactors the organization API routes from getting or deleting organizations by ID to slugs.

## Major Changes
- Refactor organization API and service
- Ensure all organization frontend uses the new API routes
- Removed API route that gets organizations by name

## Testing

Note: Some of the organization slugs have spaces in them which results in `%20` being in the URL. Even though this makes the URL a little more clunky, I still think it an improvement from using the IDs. I'm not sure if we would be able to remove it, but just wanted to leave a note so we could discuss it further.

> *This PR resolves #53*


atoneyd:
I think we should replace the spaces with dashes or something to prevent the %20 for cleaner URLs

ajaygandecha (approved):
LGTM @briannata, amazing job on this! Sorry for the delay in getting this review out. All APIs and routes are working for me. Ready to merge! Note: We may want to change the dev data in the future to include better slug conventions. Names should be lowercase, and spaces should be replaced with dashes.

PR #62, opened 2023-09-08
Update api docs

atoneyd:
### Major Changes:
This pull request adds a description for the Organization API routes, moves the Admin routes to the bottom of the page, and re-orders the demo data to be in alphabetical order.

### Testing
Ran `honcho start` and visited the /docs route to make sure the changes worked.

### DIFF
--- a/backend/test/services/organization/organization_demo_data.py
+++ b/backend/test/services/organization/organization_demo_data.py
@@ -13,36 +13,6 @@ __license__ = "MIT"
 
 # Sample Data Objects
 
-cads = Organization(
-    name="Carolina Analytics & Data Science Club",
-    slug="CADS",
-    logo="https://raw.githubusercontent.com/briannata/comp423_a3_starter/main/logos/cads.png",
-    short_description="Provides students interested in Data Science opportunities to grow.",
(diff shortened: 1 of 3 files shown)


backend/test/services/organization/organization_demo_data.py
```diff
@@ -268,7 +268,7 @@
     public = False,
 )
 
-organizations = [cads, cssg, appteam, acm, bit, arvr, ctf, enablingtech, esports, gamedev, gwc, hacknc, ktp, pearlhacks, pm, queerhack, wics]
+organizations = [appteam, acm, bit, cads, carvr, cssg, ctf, enablingtech, esports, gamedev, gwc, hacknc, ktp, pearlhacks, pm, queerhack, wics]
```
ajaygandecha, line 271: Nit: Looks good! Thanks for catching this -- I think this should be a functionality of the service, however, in the future! Data will not necessarily get added regularly, however we will not be able to manually sort production data upon each post request. A sort pipe will probably work in the frontend for this. It would be a design choice on whether we want the API to return a sorted result or not. I am thinking this can be reserved for the front end but we can discuss this @KrisJordan  / @briannata  / @jadekeegan.

ajaygandecha (approved):
LGTM, great job @atoneyd! Ready to merge despite the nit I added. We may need to make another ticket for that. Also, it would be good to get this merged in before we work on the rework of the APIs.

Issue #52, opened 2023-09-02
Migrate Organization Admin Page


jadekeegan:
# Migrate Organization Admin Page
Add create , edit, and delete functionality back to the organization admin page.

## Why?
This is important to finalize the organization feature by allowing creation, updating, and deletion of organizations.

## Acceptance Criteria
- [ ] Add create button and functionality.
- [ ] Add delete button and functionality.

PR #63, opened 2023-09-10
Migrate Organization Admin Page

jadekeegan:
This PR updates the frontend to include a tab for "Organizations" on the admin page as well as an organization creation form. These two features allow for the creation, deletion, and viewing of organizations in the database.

## Major Changes
- Added relevant TS, HTML, and CSS files for the Organization Admin page.
- Added relevant TS, HTML, and CSS files for the Organization Editor page.
- Modified modules/routing to accommodate for the new components.

## Testing

> *This PR resolves issue #52*

### DIFF
--- a/frontend/src/app/organization/organization-editor/organization-editor.component.ts
+++ b/frontend/src/app/organization/organization-editor/organization-editor.component.ts
@@ -1,0 +1,136 @@
+/**
+ * The Organization Editor Component allows organization managers to edit information
+ * about their organization which is publically displayed on the organizations page.
+ * 
+ * @author Ajay Gandecha, Jade Keegan, Brianna Ta, Audrey Toney
+ * @copyright 2023
+ * @license MIT
+ */
+
+import { Component } from '@angular/core';
+import { ActivatedRoute, Route, Router } from '@angular/router';
+import { FormBuilder, Validators } from '@angular/forms';
+import { MatSnackBar } from '@angular/material/snack-bar';
+import { Observable } from 'rxjs';
+import { profileResolver } from 'src/app/profile/profile.resolver';
+import { PermissionService } from 'src/app/permission.service';
+import { Organization, OrganizationService } from '../organization.service';
+import { Profile } from 'src/app/profile/profile.service';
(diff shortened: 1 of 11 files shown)


frontend/src/app/organization/organization-editor/organization-editor.component.ts
```diff
@@ -0,0 +1,136 @@
...
+      youtube: "",
+      heel_life: "",
+      public: false
+    };
+  }
+
+  /** Event handler to handle submitting the Update Organization Form.
+   * @returns {void}
+   */
+  onSubmit = () => {
```
ajaygandecha, line 109: We may be able to consider adding the slug generation that @KrisJordan was talking about for new organizations. The process would be as follows:
- Make string lowercase
- Using regex, select all non-alphanumeric characters and convert them to a - 
- Ensure that the slug does not already exist, otherwise as a suffix number.
    - Ex) If `cads` exists, the new org's slug would be `cads-2`.

Also, this means that we can remove the slug field entirely from the form and make its value calculated.

ajaygandecha (changes requested):
Looks great so far @jadekeegan! There are just one comment regarding a UI field and one regarding slug generation. Let me know what you think here. Overall, nice work!

frontend/src/app/organization/organization-editor/organization-editor.component.ts
jadekeegan: I think Kris did say that the slugs should still be modifiable, but maybe we could auto generate an option for them based on their organization name. However, maybe we could save this for another ticket. I think there are a few more specifications we need to add to create nicer slugs, such as determining abbreviations. The current regex would just make "carolina-analytics-and-data-science" the slug.

ajaygandecha: Good point! We can hold off for this commit then!

ajaygandecha (approved):
LGTM @jadekeegan, ready to merge! Great job!! We can figure out slugs in a later PR.

PR #66, opened 2023-09-12
Fix API doc tag

atoneyd:
Small fix to tags that got mis-edited in another PR


jadekeegan (approved):
LGTM! Thanks for noticing this issue and fixing it so quickly.

PR #67, opened 2023-09-12
Fix changes after change detection cycle error

KrisJordan:
When navigating between areas of the site there were previously a significant number of "[change made after detect](organization-page.component.html:12 ERROR Error: NG0100: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'false'. Current value: 'true'. Find more at https://angular.io/errors/NG0100)" errors.

This has to do with the interplay of the navigation component, the HTTP intercepter updating state based on the status of async HTTP requests, and the change detection cycles of angular. The solution presented in this PR feels more like an acceptable workaround, delaying state update pushes to the next event cycle tick, but could use further investigation in the future. There do not seem to be any significant downsides to this workaround besides muddying the navigation service a bit.

### DIFF
--- a/frontend/src/app/navigation/navigation.service.ts
+++ b/frontend/src/app/navigation/navigation.service.ts
@@ -19,22 +19,37 @@ export class NavigationService {
   private _error: BehaviorSubject<string | null> = new BehaviorSubject<string | null>(null);
   public error$ = this._error.asObservable();
 
-  constructor() {}
+  constructor() { }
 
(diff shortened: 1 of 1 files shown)


KrisJordan:
Incorrectly formed PR, not sure what happened here. Closing.

KrisJordan:
Fixed the PR, should be OK to merge now.

PR #70, opened 2023-09-12
Update OrganizationEntity and Add alembic migration script for Organization table

KrisJordan:
This PR introduces two changes:

1. More concretely specifies a few of the properties of OrganizationEntity to enforce schema level constraints. Specifically:
- name is not nullable and defaults to empty string
- slug is not nullable and is enforced unique at the schema level
- public is not nullable and defaults to true
2. Adds an alembic upgrade/downgrade script in preparation for production deployment


ajaygandecha:
LGTM @KrisJordan! Is there anything I should to to test this before it is merged in?

PR #89, opened 2023-09-21
Improve Organization Retrieval for Organization Details Page

jadekeegan:
This PR refactors the Organization Details page to use a route resolver to retrieve the organization and its data rather than subscribing to directly within the component. It also fixes the titleResolver so it uses the organization from the route instead of making its own HTTP request to get the organization name.

# Major Changes
- Updated the Organization Service to remove profile retrieval since the profile isn't actually used in that service.
- Added an Organization Resolver TS file to store the `organizationResolver` ResolveFn.
- Refactored the Organization Details Component to use the `organizationResolver` rather than directly calling the service and subscribing to the returned Observable.
- Changed the `title` in the route to be a child of the main route so it can inherit from the Organization Details Component parent route.
- Updated the `titleResolver` ResolveFn to use the parent data to get the organization.

# Testing
All changes were tested thoroughly on the local server by running `honcho start`.

# Developer Notes
Currently, the `titleResolver` function is in the Organization Details component. To match the other resolve functions, we could move it to its own TS file.

### DIFF
--- a/frontend/src/app/organization/organization-details/organization-details.component.ts
+++ b/frontend/src/app/organization/organization-details/organization-details.component.ts
@@ -7,31 +7,19 @@
  * @license MIT
  */
 
-import { Component, inject } from '@angular/core';
+import { Component } from '@angular/core';
 import { ActivatedRoute, ActivatedRouteSnapshot, ResolveFn, Route } from '@angular/router';
 import { MatIconRegistry } from '@angular/material/icon';
 import { DomSanitizer } from '@angular/platform-browser';
-import { map } from 'rxjs';
 import { MatSnackBar } from '@angular/material/snack-bar';
 import { profileResolver } from '/workspace/frontend/src/app/profile/profile.resolver';
-import { Organization, OrganizationService } from '../organization.service';
+import { Organization } from '../organization.service';
 import { Profile } from '/workspace/frontend/src/app/profile/profile.service';
 import { organizationResolver } from '/workspace/frontend/src/app/organization/organization.resolver'
 
 let titleResolver: ResolveFn<string> = (route: ActivatedRouteSnapshot) => {
-  let organizationSlug = route.params['slug'];
-
-  let organizationDetailSvc = inject(OrganizationService);
-  let organization$ = organizationDetailSvc.getOrganization(organizationSlug);
-  return organization$.pipe(map(organization => {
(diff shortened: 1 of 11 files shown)


briannata (approved):
LGTM! You really cleaned up the code and everything looks good on my end

frontend/src/app/organization/organization-editor/organization-editor.component.ts
```diff
@@ -7,7 +7,7 @@
  * @license MIT
  */
 
-import { Component } from '@angular/core';
+import { Component, inject } from '@angular/core';
```
atoneyd, line 10: I think this inject import is unused

frontend/src/app/organization/organization-editor/organization-editor.component.ts
```diff
@@ -27,7 +27,7 @@ export class OrganizationEditorComponent {
     path: ':slug/edit',
     component: OrganizationEditorComponent,
     title: 'Organization Editor',
-    resolve: { profile: profileResolver }
+    resolve: { profile: profileResolver}
```
atoneyd, line 30: Looks like a space was accidentally removed after profileResolver

frontend/src/app/organization/organization-details/organization-details.component.ts
```diff
@@ -7,30 +7,19 @@
...
 import { ActivatedRoute, ActivatedRouteSnapshot, ResolveFn, Route } from '@angular/router';
 import { MatIconRegistry } from '@angular/material/icon';
 import { DomSanitizer } from '@angular/platform-browser';
-import { Observable, map } from 'rxjs';
 import { MatSnackBar } from '@angular/material/snack-bar';
 import { profileResolver } from '/workspace/frontend/src/app/profile/profile.resolver';
-import { Organization, OrganizationService } from '../organization.service';
-import { Profile, ProfileService } from '/workspace/frontend/src/app/profile/profile.service';
+import { Organization } from '../organization.service';
+import { Profile } from '/workspace/frontend/src/app/profile/profile.service';
```
atoneyd, line 17: Maybe a stupid question but is there a reason for us to be importing Profile/use profile on this page?

atoneyd (changes requested):
Some small nit comments but overall looks good

Issue #81, opened 2023-09-15
Clean Up Organization Backend Code


jadekeegan:
# Clean Up Backend Code
Clean up and refactor any organization code in the backend that is dead or not idiomatically correct.

## Why?
We want our code to be as readable and efficient as possible.

## Acceptance Criteria
- [ ] Edit `backend/services/organization.py` to raise a more specific exception for get_from_slug (line 77)
- [ ] Edit `backend/services/organization.py` to also check whether a user is a site admin for updating an organization (line 121-131)
- [ ] Edit `backend/services/organization.py` to change the current code on lines 92-95 to be more idiomatic (`...._scalar_one_or_none`).

PR #95, opened 2023-09-27
Clean Up Organization Backend

briannata:
This PR cleans up a couple of organization backend methods.

# Major Changes
- Made `get_from_slug` more idiomatic, using `one_or_none()` instead of `[0]`
- Changed organization update method to check for admin permissions instead of org roles

# Testing
All changes were tested thoroughly on the local server by running `honcho start`. Note: since we don't have the edit organization functionality yet, I wasn't able to fully test whether my admin permission check worked.

Resolves issue #81

### DIFF
--- a/backend/services/organization.py
+++ b/backend/services/organization.py
@@ -60,8 +60,8 @@ class OrganizationService:
 
         # Checks if the organization already exists in the table
         if organization.id:
-            # If so, raise an error
-            raise Exception(f"Duplicate organization found with ID: {organization.id}")
+            # Set id to None so database can handle setting the id
+            organization.id = None
 
         else:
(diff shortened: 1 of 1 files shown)


ajaygandecha (approved):
LGTM! Ready to merge - great work!

atoneyd:
Can you address the following comment from Kris in this ticket:

> Non-blocking, but it seems surprising that we would have an organization id at all here; I believe we should likely rely upon the auto-assigned IDs. If an ID is provided to create, we should either ignore it or error out. Minor nit to touch up post deploy.

This is regarding the create method under backend/services/organization.py

atoneyd (changes requested):
nit comment added

PR #97, opened 2023-09-30
Add alembic migration to add shorthand column

KrisJordan:
Here's the migration to add the shorthand column to the production database. The one implementation decision of note is that because this table already exists with populated data, the upgrade process is:

1. Add column (without non-null constraint)
2. Copy over the name column to the shorthand column (for now)
3. Add the non-null constraint

This means for orgs whose shorthand is different from their name, we'll need to either edit them via admin panel or write a separate post-deploy script if admin panel isn't ready and we really need different shorthands.


briannata (approved):
LGTM! I'll make a ticket to add an admin update organization feature so we can change all of the `shorthand`s when `organization-migration-fix` is merged in.

Issue #98, opened 2023-10-01
Update Org Frontend Functionality


briannata:
# Update Org Frontend Functionality
Add the appropriate methods and buttons to the frontend so that admins can see an edit button and successfully be navigated to the organization editor

## Why?
We need to be able to change the shorthand for organizations once the migration goes through (since it is just copying the name into the shorthand column for now).

## Acceptance Criteria
- [x] Add edit button to organization details page if user is an admin
- [x] Navigate to organization editor on button clicked

Issue #96, opened 2023-09-29
Fix Z-Index of Admin Tabs


briannata:
# Fix Z-Index of Admin Tabs
Edit the Z-Index of the tabs so that they aren't show when you scroll and they are in the same space as the blue title section

## Why?
This fix a bug on the admin component so that you don't see the admin tab names over the blue title section. It will make the page neater and clean it up.

## Acceptance Criteria
- [ ] Fix the z-index of the admin tabs so they don't show over the blue title section

Issue #85, opened 2023-09-15
Fix Backend Exceptions


ajaygandecha:
# Fix Backend Exceptions
This issue creates more specific exceptions to better describe the errors that are occurring in the backend.

## Why?
Currently, many of the exceptions are throwing as a 404 error, which in many cases is either not accurate or representative of the actual error. We need to create custom exceptions to fix this.

## Acceptance Criteria
- [] Use `with pytest.raises(Error)` instead of using try/except blocks in the unit tests.

Issue #83, opened 2023-09-15
Improve UX When Navigating to an Invalid Organization


jadekeegan:
# Improve UX When Navigating to an Invalid Organization
Post an alert/dialog to let the user know that the organization is not valid and then redirect to the main `/organizations` page on alert dismissal.

## Why?
This is important to improving user interaction with the Organizations/Organization Details pages.

## Acceptance Criteria
- [ ] Create a dialog component that emits a message indicating that the organization does not exist.
- [ ] Ensure the new component contains a button to dismiss the pop up.
- [ ] Navigate to the `/organizations` page after clicking the dismiss button.

PR #99, opened 2023-10-03
Organization Fixes

KrisJordan:
This PR rolls up a number of fixes for the Organization feature of the XL web application.


briannata:
@KrisJordan Hi Kris, I think we still have a few issues that we want to finish before merging this pull request.

- Issue #98 would allow us to update the `shorthand`s so they are not the same as the names
- Issue #96 fixes the admin tab z-index bug
- Issue #85 would make more specific/correct exceptions for our organization backend
- Issue #83 improves the UX when navigating to an invalid organization

I think we could definitely do these later, though, and I can finish up Issue #98 by tomorrow night so we can merge that one in and have the ideal `shorthand`s for our production data.

There are a few other issues that could be merged with this branch, but I'm not sure if we have enough time to finish them before this Friday.

PR #102, opened 2023-10-06
Fix Organization List HTML Update After Delete

jadekeegan:
This PR implements a quick fix to ensure that the HTML updates after an organization is deleted on the organization admin page.

## Major Changes
- Added a line in the subscription to deleteOrganization() to filter out the organization from the component's organizations array.

## Testing
All changes were tested on the local server after running `honcho start`.

## Developer Notes
This is a temporary fix (modeled after how the admin roles list is updated). This will be changed if the code is refactored to use RXJS objects throughout the application.


ajaygandecha:
Hey @jadekeegan may need to change base branch!

ajaygandecha (approved):
LGTM @jadekeegan !

PR #107, opened 2023-10-14
Implement the Events Feature Backend

ajaygandecha:
This major pull PR provides the implementation of the Python backend for the Events feature. This includes the SQLAlchemy Entities and Pydantic models, backend services, FastAPI routes, and test data. 

- Added the `EventEntity` SQLAlchemy Entity to model how events should be stored in the PostgreSQL database.
- Established a one-to-many relationship between the `organizations` and `events` tables through the SQLAlchemy Entities.
- Added the `Event` and `EventDetail` Pydantic models to represent the shape of events data to be exposed by the APIs. The `EventDetail` model exists to prevent circularity between the `Event` and existing `Organization` Pydantic models.
### Scripts Changes
- Added demo and test data for events. This data should populate when the `demo_reset.py` or `test_reset.py` scripts are run respectively.
- Created the `date_maker` helper function to ensure that the dates for demo events data are generated based on when the reset script is run. No more changing date constants!
### API Changes
Exposes the following APIs to the Angular Frontend via FastAPI:
- GET `api/events`: Returns all of the events from the database.
- GET `api/events/:id`: Returns an event from the database based on its id.
 - GET `api/events/organization/:id`: Returns the events for an organization. ***NOTE:** We may want to consider deprecating this API route and use `api/organizations/` instead.*
 - POST `api/events`: Adds a new event to the database.
- PUT `api/events`: Updates an existing event in the database.
- DELETE `api/events/:id`: Deletes a specific event from the database.

*NOTE: All of the mutating API routes are locked based on user permissions.*

### Service Changes
- Implemented the `EventService` to connect the APIs to data in the database.

## Testing
These changes were thoroughly tested. First, we recreated and repopulated the database. Then, every API route was tested in the FastAPI docs page to ensure that they work correctly. I also used a SQL database inspector to ensure that the correct changes and relationships were being reflected in the database.

## Future Considerations
- @atoneyd: Write pytests to ensure that the `EventService` works in all situations.
- Consider deprecating the  "GET `api/events/organization/:id`" API route, as discussed above.
- Discuss how we want to handle event durations / end times in the database. Right now, based on work done last semester, we are only keeping track of the start time. This may be a reasonable band-aid fix in the short term, since event durations are *generally* assumed to be 1hr unless specified otherwise for organization events (not departmental events).
- Once this pull request is approved, I (@ajaygandecha) will integrate this with the frontend that we have.
- We may need to consider changing the permission flag used to determine whether events can be created or not. We may actually need to tie this to whether or not a user has admin permission *for the organization the event will be for*. @KrisJordan thoughts?

### DIFF
--- a/backend/services/event.py
+++ b/backend/services/event.py
@@ -7,6 +7,7 @@ from ..database import db_session
 from backend.models.event import Event
 from backend.models.event_details import EventDetails
 from ..entities import EventEntity
+from .permission import PermissionService
 
 class EventNotFoundException(Exception):
     """EventNotFoundException is raised when trying to access an event that does not exist."""
@@ -21,9 +22,10 @@ class EventService:
     # Current SQLAlchemy Session
     _session: Session
 
-    def __init__(self, session: Session = Depends(db_session)):
+    def __init__(self, session: Session = Depends(db_session), permission: PermissionService = Depends(),):
         """Initializes the `EventService` session"""
         self._session = session
+        self._permission = permission
 
     def all(self) -> list[EventDetails]:
         """
@@ -37,9 +39,9 @@ class EventService:
         entities = self._session.scalars(query).all()
 
         # Convert entries to a model and return
-        return [entity.to_model() for entity in entities]
+        return [entity.to_details_model() for entity in entities]
 
-    def create(self, subject: User, event: Event) -> EventDetails:
+    def create(self, subject: User, event: EventDetails) -> EventDetails:
         """
         Creates a event based on the input object and adds it to the table.
         If the event's ID is unique to the table, a new entry is added.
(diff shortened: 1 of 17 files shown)


jadekeegan:
Regarding the "GET `api/events/organization/:id`" route, I believe we should change it to `api/organizations/:slug/events` or something similar instead to ensure that we have consistency in the field we are using to access Organizations.

I think we definitely need to add a field for end time!

As for permissions, I believe Kris mentioned that adding something like `organizations/:slug` to the permission value should do the trick!

backend/services/event.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # Current SQLAlchemy Session
+    _session: Session
+
+    def __init__(self, session: Session = Depends(db_session), permission: PermissionService = Depends()):
+        """Initializes the `EventService` session"""
+        self._session = session
+        self._permission = permission
+
+    def all(self) -> list[EventDetails]:
```
KrisJordan, line 30: This will ultimately need to be more nuanced, right -- like a guarded by a date range or some other form of pagination? Thinking about what the use case is for this method five years from now, with an assumption that we will not delete old events.

backend/services/event.py
```diff
@@ -0,0 +1,165 @@
...
+        If the event's ID already exists in the table, raise an exception.
+
+        Parameters:
+            subject: a valid User model representing the currently logged in User
+            event: a valid Event model representing the event to be added
+
+        Returns:
+            EventDetails: a valid EventDetails model representing the new Event
+        """
+        self._permission.enforce(subject, "events.create", f"events")
```
KrisJordan, line 57: This permission could be on "organization.events.create" "organization/{org}"

backend/services/event.py
```diff
@@ -0,0 +1,165 @@
...
+        Update the event
+        If none found, a debug description is displayed.
+
+        Parameters:
+            event: a valid Event model
+
+        Returns:
+            EventDetails: a valid EventDetails model representing the updated event object
+        """
+        self._permission.enforce(subject, "events.update", f"events")
```
KrisJordan, line 124: Same comment here, agree this could be based on org.events permission for an entire org.

KrisJordan (approved):
LGTM with a few permission suggestions but non-blocking.

PR #108, opened 2023-10-17
Implement the Frontend for the Events Feature

ajaygandecha:
This major pull PR provides the implementation of the Angular frontend for the Events feature. This includes all components, widgets, services, resolvers, and helper TypeScript data models. This feature will ultimately enable students on the CSXL site to explore events hosted by CS Department-affiliated organizations on campus.

### Events Page
- Add the events page at `/events` to showcase all events hosted by CS Organizations at UNC. On desktop, the events page features a dual-column setup with a list of events on the left and details about the event on the right. On mobile, only the list displays. There is an events detail page that opens when a user taps on an event in mobile.
- On desktop, the list of events is universally scrollable; the cursor can be anywhere on the page and only that section of the page scrolls. The right side containing events details sticks to the same position on the page.
- The events page groups events based on their date.
- The events page also has a search bar that queries over event names, descriptions, and organization names. The search bar is also sticky and stays at the top of the page.
- The events page is fully responsive, both to screen sizes and to light / dark mode.
- Users can press on the "share" button on an event to copy a permalink to the event to their clipboard.
- Users are also able to navigate to hosting organizations when exploring the details of events.

### Events Module and Services
- All business logic for events are abstracted in to the events service. The events module also decouples the code for events from the rest of the application, lessening resources needed to load the events page and therefore may help on performance. This same approach was used for organizations.
- This module, like the Organizations module previously, uses *resolvers* to pass in events data. This seems clean, simple, and effective.

## Testing
These changes were thoroughly tested. First, we recreated and repopulated the database. Then, all features were tested on the frontend by running `honcho start`.

## Future Considerations
- Currently, there is no admin functionality to add, edit, or delete events. This needs to still be implemented. (@atoneyd / @briannata / @jadekeegan )
- There is a button commented-out that would serve as a way for users to download calendar events directly to their computer. This would be a great feature, but likely good to be saved for a future commit.
- We need to handle the situation where there are no events. The page works in this instance, however we may want a different UI than what we have right now.
- @KrisJordan, this is a design choice to discuss, but would it be beneficial to *always* have the current date at the top, even if there are no events?
- We may want to add some sort of pagination / load limit to this. This would be a decent ticket since it has a large effect on both the frontend and the backend. I do not anticipate there would be so many events that this would happen (for future events), however this may be a concern for past events when shown on the organizations pages.


jadekeegan:
It's looking great Ajay!! A few things I noticed:

Most important, whenever I click on anything in the page, it causes the following console error:
image

As for some more minor nits:
1. When you scroll down on the events list, there is a small line between the title and the search bar where you can see a little bit of the Events list text.
image

2. Could we move the organization name to the bottom of the detail card and see how that looks? The card just feels a bit bottom heavy right now since all the shorter text items are at the top and the long description is at the bottom.

3. Could we give a little more screen width to the Events list? The list feels quite stacked, and there is definitely room to make the Event details cards smaller since they all have white space at the bottom.

4. Could we use short descriptions in the Events list? Right now it looks like the items in the Events list have pretty much the exact same information as the Event Details card/page, which makes the extra pages a little redundant. It also makes the Events list a little harder to scroll through quickly.

5. Can we change the button for navigating to the specific organization to have a pointer clicker on hover? It highlights right now, but the cursor remains the same.

jadekeegan (changes requested):
It's looking great Ajay!! A few things I noticed:

Most important, whenever I click on anything in the page, it causes the following console error:
image

As for some more minor nits:
1. When you scroll down on the events list, there is a small line between the title and the search bar where you can see a little bit of the Events list text.
image

2. Could we move the organization name to the bottom of the detail card and see how that looks? The card just feels a bit bottom heavy right now since all the shorter text items are at the top and the long description is at the bottom.

3. Could we give a little more screen width to the Events list? The list feels quite stacked, and there is definitely room to make the Event details cards smaller since they all have white space at the bottom.

4. Could we use short descriptions in the Events list? Right now it looks like the items in the Events list have pretty much the exact same information as the Event Details card/page, which makes the extra pages a little redundant. It also makes the Events list a little harder to scroll through quickly.

5. Can we change the button for navigating to the specific organization to have a pointer clicker on hover? It highlights right now, but the cursor remains the same.

ajaygandecha:
@jadekeegan 

> It's looking great Ajay!! A few things I noticed:
> 
> Most important, whenever I click on anything in the page, it causes the following console error: image

Noticed this too - want to ping @KrisJordan here to see if he has any suggestions on this implementation, since it seemed quite hacked to me! Is the date grouping something that should be moved to the backend or is it fine this way? Second, do we need to use `RxObject` here?


> As for some more minor nits:
> 
> 1. When you scroll down on the events list, there is a small line between the title and the search bar where you can see a little bit of the Events list text.
> 
> image

Fixed!

> 2. Could we move the organization name to the bottom of the detail card and see how that looks? The card just feels a bit bottom heavy right now since all the shorter text items are at the top and the long description is at the bottom.

We need to discuss this further! The reason why this implementation seems the way it does is because the demo data does not contain a lot of text. In the future, it is very possible that event descriptions are 1-3 paragraphs long - far longer than what I have added. When it is longer, this effect is not as drastic. In addition, generally most post UIs often have the "author" at the top of the page - I think in this case, organizations are the authors of events. 

> 3. Could we give a little more screen width to the Events list? The list feels quite stacked, and there is definitely room to make the Event details cards smaller since they all have white space at the bottom.

Yup, we can do this! Let me know if there is a certain pixel amount you prefer. This may also be due to screen size differences also so definitely send me a screenshot.

> 4. Could we use short descriptions in the Events list? Right now it looks like the items in the Events list have pretty much the exact same information as the Event Details card/page, which makes the extra pages a little redundant. It also makes the Events list a little harder to scroll through quickly.

I added truncation to 3 lines here already, but as mentioned earlier, the demo data descriptions are not long enough to properly showcase this! If we want to add a short description field to the entity, we can as well.

> 5. Can we change the button for navigating to the specific organization to have a pointer clicker on hover? It highlights right now, but the cursor remains the same.

Done.

frontend/src/app/event/event-details/event-details.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  selector: 'app-event-details',
+  templateUrl: './event-details.component.html',
+  styleUrls: ['./event-details.component.css']
+})
+export class EventDetailsComponent {
+
+  /** Route information to be used in App Routing Module */
+  public static Route = {
+    path: ':id',
+    title: 'Event Detail - TODO: Make Title Dynamic',
```
KrisJordan, line 32: Seems like the title resolver is setup but ignored?

frontend/src/app/event/widgets/event-card/event-card.widget.html
```diff
@@ -0,0 +1,16 @@
...
+    <div style="display: flex; flex-direction: column; padding: 12px;">
+        <div style="display: flex; flex-direction: row;">
+            <p style="font-weight: 500; margin: 0px; width: 75%;">{{event.name}}</p>
+            <p style="margin: 0px; text-align: end; width: 25%;">{{event.time | date:'shortTime'}}</p>
+        </div>
+        <div style="display: flex; flex-direction: row; margin-top: 8px;">
+            <p style="margin: 0px; width: 50%;">{{event.organization.shorthand}}</p>
+            <p style="margin: 0px; text-align: end; width: 50%;">{{event.location}}</p>
+        </div>
+        <p class="description-text">{{event.description}}</p>
```
KrisJordan, line 14: I wonder if we should drop the description on the card or have a very short version of it (perhaps first line only?). It's currently unclear what benefit exists in selecting an event from the main list at the moment.

frontend/src/app/event/widgets/event-detail-card/event-detail-card.widget.ts
```diff
@@ -0,0 +1,30 @@
...
+})
+export class EventDetailCard {
+
+    @Input() event!: Event
+
+    constructor(protected snackBar: MatSnackBar) { }
+
+    onShareButtonClick() {
+        navigator.clipboard.writeText("https://csxl.unc.edu/events/" + this.event.id);
+        this.snackBar.open("Event link copied to clipboard.", "", {duration: 3000})
```
KrisJordan, line 28: This is cool and a really nice touch.

KrisJordan: Maybe overkill, but perhaps worth using the front-end's host rather than production. I was momentarily confused why I could navigate to an event in mobile view, but copy/paste was failing lol

KrisJordan (approved):
This is looking good. I don't see any significant blockers in making progress.

One thing I noticed in trying it locally was that on my desktop when the view port was longer than the list of events, something different happened with the gutter between the cards. Image attached below.

Additionally, visually, I wonder if the details card should not attempt to fill vertically if we're going out of the way for a desktop experience (imagining this with a single event toward the end of a semester):
 
Screenshot 2023-10-17 at 8 43 24 PM

Finally, I'm still not sure we've landed in the right final solution on time and am thinking start/end may make sense (and just displaying start and end time rather than start + duration) for the various kinds of events (6pm-7pm for a lecture, 12am-11:59pm for a hack day). I don't know that this is blocking and will be easy to add/change later, but came to mind.

frontend/src/app/event/event-page/event-page.component.ts
```diff
@@ -0,0 +1,125 @@
...
+      let dateString = this.datePipe.transform(event.time, 'EEEE, MMMM d, y') ?? ""
+      // Add the event
+      let newEventsList = groups.get(dateString) ?? []
+      newEventsList.push(event)
+      groups.set(dateString, newEventsList)
+    })
+    
+    // Update the data
+    this.eventsPerDay = groups;
+  }
```
KrisJordan, line 109: A more idiomatic implementation of this method would be to make it pure and return the groups than assign the mutation from within (the assignment would happen in the constructor and it makes the constructor a little more legible to know what's going on. Additionally, you can avoid assigning an empty Map placeholder and still avoid the not initialized type check.

frontend/src/app/event/event-page/event-page.component.html
```diff
@@ -0,0 +1,55 @@
...
+                <mat-card class="events-section" appearance="outlined" style="padding: 16px;">
+
+                    <div *ngFor="let date of eventsPerDay.keys()" style="padding-bottom: 16px;">
+                        <mat-card-title>{{date}}</mat-card-title>
+                        <mat-divider style="padding-bottom: 8px;" />  
+                        <event-card *ngFor="let event of eventsPerDay.get(date)" [event]="event"
+                        [disableLink]="innerWidth > 900" [selected]="event === selectedEvent"
+                        (onClicked)="onEventCardClicked($event)"
+                        />  
+                    </div>
```
KrisJordan, line 20: I believe this is the section causing the change after checked and a more idiomatic way to loop over key/value in map is: https://angular.io/api/common/KeyValuePipe

KrisJordan (changes requested):
I am hopeful the suggested change will avoid the errors in navigating front-end.

frontend/src/app/event/widgets/event-card/event-card.widget.html
ajaygandecha: @KrisJordan I am actually in support of just dropping it for right now! I agree that there is a bit of a redundancy here. I will commit this change, and we can always add it again later if needed.

frontend/src/app/event/event-page/event-page.component.html
ajaygandecha: This worked!!!

jadekeegan (approved):
A few small nits, but nothing blocking:
Could we remove the line at the bottom or add extra padding to make it balanced with the padding at the top?
image

The Events card goes past the Search Bar right now.
image

Could we use the full organization names instead of the slugs in the Event Details card?

PR #113, opened 2023-10-20
Add tests for events backend

atoneyd:
This PR adds tests for the EventService in the backend

## Major Changes
- Add tests for every method in the EventService class
- Modify some of the methods to take in an `Event` object rather than `EventDetails`, after refactoring `Event` object to include and organization_id (this was changed after group discussion, we weren't sure why it took in EventDetails in the first place)

## Testing
Ran pytest, made sure frontend still works although I did not test the events tab because it does not exist right now.


backend/test/services/event/event_test.py
```diff
@@ -0,0 +1,126 @@
...
+def test_create_event_as_user(event_svc_integration: EventService):
+    """Test that any user is *unable* to create new events."""
+    with pytest.raises(UserPermissionException):
+        event_svc_integration.create(user, to_add)
+        pytest.fail()  # Fail test if no error was thrown above
+
+
+def test_get_events_from_organization(event_svc_integration: EventService):
+    """Test that list of events can be retrieved based on specified organization."""
+    event_svc_integration.create(root, to_add)
```
ajaygandecha, line 79: Small Nit / Discussion: Unsure if we should have a create call in here since this could be implemented incorrectly - but since the unit tests run sequentially, it may be okay.

backend/test/services/event/event_test_data.py
```diff
@@ -15,36 +15,58 @@
...
-event_two = EventEntity(
+event_two = Event(
+    id=2,
     name="CS+SG Workshop",
     time=date_maker(days_in_future=2, hour=19, minutes=0),
     location = "SN 014",
     description="This is a sample description.",
     public=True,
-    organization_id=cssg.id
+    organization_id=2
```
ajaygandecha, line 35: Same as above

ajaygandecha (changes requested):
Looks great @atoneyd, amazing work!! Left some small nits but should be a quick fix! Then, I'll go ahead and work on resolving the merge conflicts. Feel free to re-request when ready!

backend/test/services/event/event_test_data.py
```diff
@@ -15,36 +15,58 @@
...
-event_one = EventEntity(
+event_one = Event(
+    id=1,
     name="Carolina Data Challenge",
     time=date_maker(days_in_future=1, hour=10, minutes=0),
     location="Sitterson Hall Lower Lobby",
     description="Mark your calendars for the 2023 Carolina Data Challenge (CDC)! CDC is UNC's weekend-long datathon that brings together hundreds of participants from across campus, numerous corporate sponsors, tons of free food as well as merch, and hundreds of dollars of prizes!",
     public=True,
-    organization_id=cads.id
+    organization_id=1
```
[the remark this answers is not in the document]
atoneyd: I did this because I ran into an error doing cads.id due to the fact that id can be int | None. I can add a conditional to see if it exists if you think it's better to still do it by calling the object.

backend/test/services/event/event_test.py
atoneyd: Yeah I figured since the test above tests creating an object it would be fine but I can refactor if necessary.

backend/test/services/event/event_test_data.py
atoneyd: I'll try to resolve these tomorrow

backend/test/services/event/event_test.py
ajaygandecha: @atoneyd no worries, I think this is okay!

backend/test/services/event/event_test_data.py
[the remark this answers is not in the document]
ajaygandecha: @atoneyd  No worries - you may be able to try `cads.id!` and see that helps, otherwise we can leave it as hardcoded numbers.

ajaygandecha (approved):
@atoneyd LGTM! Ready to merge - great work!

PR #114, opened 2023-10-21
Add Events to the CSXL Application

ajaygandecha:
# The Events Update

This update adds events to the CSXL application. Students will now be able to explore the rich variety of events hosted by the CS Department and affiliated organizations, as well as search for events they may be interested in. Leaders of organizations are also able to add events for their organizations.

### Events Landing Page 

The Events page is the landing page for students to explore all CS events. Students may click on event tiles to learn more about each event, or use the search bar to filter available options by name, description, or hosting organizations.

### Events Detail Page

On mobile, clicking on events in the Events page navigates to this event details page. In addition, permalinks to organizations will redirect to this page.

### Mobile Responsiveness
The events feature is designed to be user-friendly on both desktop and mobile devices. On mobile, the events page contains a scrollable list of events. On desktop, event details are displayed on the right sidebar.

### Minor Improvements
- Pages may load slightly faster.
- The codebase is better documented in preparation for COMP 423.
- All APIs have been documented and are accessible on the `/docs` page.

### Testing
- All backend and frontend features have been extensively tested over the last few weeks. In addition, unit tests have been added to extensively test the backend services that handle dealing with events.

### Attributions
This feature was developed as part of COMP 423 by final group A3 - @ajaygandecha, @jadekeegan, @briannata, and @atoneyd - with significant contributions and guidance from @KrisJordan.


ajaygandecha:
Note: Noticing issue with sorting of events. Commit will be pushed to fix this soon.

KrisJordan:
I'll be able to gen the migration, no problem. Looks like we will have some conflicts to resolve in rebasing onto main, I think that can happen just before merge. However, if we do wait until that point, I'll recommend waiting on changing the styles everywhere else and doing that as it's own separate branch and pr that lands after this one. We can still use the new styles in the events scope of this PR, as planned.

backend/api/events.py
```diff
@@ -0,0 +1,145 @@
...
+        HTTPException 404 if delete() raises an Exception
+    """
+   
+    try:
+        # Try to delete event
+        event_service.delete(subject, id)
+    except (EventNotFoundException, UserPermissionException) as e:
+        # Raise 404 exception if search fails
+        # - This would occur if there is no response or if item to delete does not exist
+        raise HTTPException(status_code=404, detail=str(e))
```
KrisJordan, line 145: Comments again redundant. Here, I think the right move for `UserPermission` is just to let the application-level middleware handle it.

backend/entities/event_entity.py
```diff
@@ -0,0 +1,106 @@
...
+            EventDetails: An EventDetails model for API usage.
+        """
+        return EventDetails(
+            id=self.id, 
+            name=self.name, 
+            time=self.time, 
+            location=self.location, 
+            description=self.description, 
+            public=self.public,
+            organization_id=self.organization_id,
```
KrisJordan, line 104: For `EventDetails`, it seems like we should drop the `organization_id` field and let it reside in the nested `organization` object. This will likely touch a few places, but seems like the right way of climbing the abstraction ladder.

backend/entities/organization_entity.py
```diff
@@ -43,16 +47,17 @@ class OrganizationEntity(EntityBase):
     heel_life: Mapped[str] = mapped_column(String)
     # Whether the organization can be joined by anyone or not
     public: Mapped[bool] = mapped_column(Boolean, nullable=False, default=True)
-
+    
+    # NOTE: This field establishes a one-to-many relationship between the organizations and events table.
+    events: Mapped[list['EventEntity']] = relationship(back_populates="organization", cascade="all,delete")
```
KrisJordan, line 52: Is `cascade="all,delete"` redundant in the `delete`?

backend/entities/organization_entity.py
```diff
@@ -95,4 +100,29 @@ def to_model(self) -> Organization:
...
+            short_description=self.short_description,
+            long_description=self.long_description,
+            website=self.website,
+            email=self.email,
+            instagram=self.instagram,
+            linked_in=self.linked_in,
+            youtube=self.youtube,
+            heel_life=self.heel_life,
+            public=self.public,
+            events=[event.to_model() for event in self.events]
```
KrisJordan, line 127: Something we'll need to think about: do we actually want `events` to be a part of an `OrganizationDetails` model or should we just have a separate end point for querying the events of a specific org? My concern is looking toward the future when past events outnumber upcoming events: it seems this is getting all historical events when that's unlikely to be what shows in the page or of interest to the user.

backend/migrations/versions/63fc48273e15_add_events_table.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('event',
+        sa.Column('id', sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column('name', sa.String(), nullable=False),
+        sa.Column('time', sa.DateTime(), nullable=False),
+        sa.Column('location', sa.String(), nullable=False),
+        sa.Column('description', sa.String(), nullable=False),
```
KrisJordan, line 26: @KrisJordan needs to look into whether this column type is correct. This needs to be a SQL Text field with arbitrary length available.

backend/models/event.py
```diff
@@ -0,0 +1,22 @@
...
+    of the `Event` database in the PostgreSQL database
+    """
+    
+    id: int | None=None
+    name: str
+    time: datetime
+    location: str
+    description: str
+    public: bool
+    organization_id: int
```
KrisJordan, line 22: I'm not certain this makes sense to have in the `Event` model. This feels beneath the model level of abstraction.

backend/models/event_details.py
```diff
@@ -0,0 +1,13 @@
...
+class EventDetails(Event):
+    """
+    Pydantic model to represent an `Event`, including back-populated
+    relationship fields.
+
+    This model is based on the `EventEntity` model, which defines the shape
+    of the `Event` database in the PostgreSQL database.
+    """
+    
+    organization: Organization | None = None
```
KrisJordan, line 13: Should this actually be `| None`?  Is there a case when an event has no organization affinity?

KrisJordan: (And if there is... how do we add and edit them :)

backend/models/user.py
```diff
@@ -29,11 +36,23 @@ class User(UserIdentity, BaseModel):
...
 
 class ProfileForm(BaseModel):
-    """A profile form is a form for updating a user's profile."""
+    """
+    Pydantic model to represent fields for a form when updating
+    a user profile on the frontend.
+
+    This model is based on the `UserEntity` model, which defines the shape
+    of the `User` database in the PostgreSQL database
+    """
```
KrisJordan, line 55: Nice work adding commentary throughout!

backend/services/event.py
```diff
@@ -0,0 +1,176 @@
...
+            event: a valid Event model representing the event to be added
+
+        Returns:
+            EventDetails: a valid EventDetails model representing the new Event
+        """
+        
+        # Ensure that the user has appropriate permissions to create users
+        self._permission.enforce(subject, "organization.events.manage", f"organization/{event.organization_id}")
+
+        # Checks if the role already exists in the table
```
KrisJordan, line 59: s/role/event/

backend/services/event.py
```diff
@@ -0,0 +1,176 @@
...
+
+        Parameters:
+            id: a valid int representing a unique event ID
+
+        Returns:
+            Event: Object with corresponding ID
+        """
+
+        # Query the event with matching id
+        entity = self._session.query(EventEntity).get(id)
```
KrisJordan, line 86: I believe this is the infracting line generating warnings due to making use of SQLAlchemy 1.0 get-by-id conventions. The 2.0 idiom is: `self._session.get(EventEntity, id)`

backend/services/event.py
```diff
@@ -0,0 +1,176 @@
...
+        # Query the event with matching organization slug
+        events = self._session.query(EventEntity).filter(EventEntity.organization_id == id).all()
+        
+        # Convert entities to models and return
+        return [event.to_details_model() for event in events]
+
+    def update(self, subject: User, event: Event) -> EventDetails:
+        """
+        Update the event
+        If none found, a debug description is displayed.
```
KrisJordan, line 116: I'm a little uncertain what 'debug description' refers to.

backend/services/event.py
```diff
@@ -0,0 +1,176 @@
...
+
+        Returns:
+            EventDetails: a valid EventDetails model representing the updated event object
+        """
+        
+        # Ensure that the user has appropriate permissions to update users
+        self._permission.enforce(subject, "organization.events.manage", f"organization/{event.organization_id}")
+
+        # Query the event with matching id
+        obj = self._session.query(EventEntity).get(event.id)
```
KrisJordan, line 129: Another place to move to the newer `get` method functionality.

backend/services/event.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # Check if result is null
+        if obj:
```
KrisJordan, line 132: I wonder if it'd be more idiomatic to check `if obj is not Null` and raise the exception, otherwise let the else branch be implicit from that point and not have an `else`.

backend/services/event.py
KrisJordan: Additionally, we should use a more descriptive name than `obj`, such as `event_entity`

backend/services/event.py
```diff
@@ -0,0 +1,176 @@
...
+
+        Parameters:
+            id: an int representing a unique event ID
+        """
+        
+        # Find object to delete
+        event = self._session.query(EventEntity).get(id)
+
+        # Ensure that the user has appropriate permissions to delete users
+        self._permission.enforce(subject, "organization.events.manage", f"organization/{event.organization_id}")
```
KrisJordan, line 164: Move this check above the query for a slight efficiency gain in the scenario where the user does not have permission.

backend/services/exceptions.py
```diff
@@ -10,3 +16,16 @@ def __init__(self, action: str, resource: str):
...
+    def __init__(self, id: int | None):
+        super().__init__(
+            f'No event found with matching ID: {id}')
+
+class OrganizationNotFoundException(Exception):
+    """OrganizationNotFoundException is raised when trying to access an organization that does not exist."""
+
+    def __init__(self, id: str):
+        super().__init__(
+            f'No organization found matching slug/id: {id}')
```
KrisJordan, line 31: I would be inclined to do one of two things:

1. Simplify by simply using `ResourceNotFoundException` anywhere we're using these more specific types and remove these types entirely.
2. Change the superclass of these two types to be `ResourceNotFoundException`.

The motivation for either is because we have a top-level exception handler in FastAPI for responding 404 when a `ResourceNotFoundException` is encountered.

frontend/src/app/event/event-details/event-details.component.ts
```diff
@@ -0,0 +1,51 @@
...
+})
+export class EventDetailsComponent {
+
+  /** Route information to be used in Event Routing Module */
+  public static Route = {
+    path: ':id',
+    title: 'Event Details',
+    component: EventDetailsComponent,
+    resolve: { profile: profileResolver, event: eventDetailResolver },
+    children: [{ path: '', title: titleResolver, component: EventDetailsComponent }]
```
KrisJordan, line 35: I'm a little surprised by this `children` attribute. Is this accidental carryover from another place?

frontend/src/app/event/event-editor/event-editor.component.html
```diff
@@ -0,0 +1,31 @@
+<form [formGroup]="eventForm" (ngSubmit)="onSubmit()" *ngIf="adminPermission$ | async">
+    <mat-card>
+        <mat-card-header>
+            <mat-card-title *ngIf="event.id == -1">Create Event</mat-card-title>
+            <mat-card-title *ngIf="event.id != -1">Update Event</mat-card-title>
```
KrisJordan, line 5: Hmmmmmm. I wonder if `id` should just be `undefined` in the case of a new event? That's probably more idiomatic than using `-1`

frontend/src/app/event/event-editor/event-editor.component.ts
```diff
@@ -0,0 +1,141 @@
...
+import { Event } from '../event.model';
+
+@Component({
+  selector: 'app-event-editor',
+  templateUrl: './event-editor.component.html',
+  styleUrls: ['./event-editor.component.css']
+})
+export class EventEditorComponent {
+  public static Route: Route = {
+    path: 'organizations/:slug/events/:id/edit',
```
KrisJordan, line 32: This path is fine to me!

frontend/src/app/event/event-editor/event-editor.component.ts
```diff
@@ -0,0 +1,141 @@
...
+    this.adminPermission$ = this.permission.check('organization.events.manage', `organization/${this.organization!.id}`);
+  }
+
+  /** Event handler to handle submitting the Create Event Form.
+   * @returns {void}
+  */
+  onSubmit = () => {
+    if (this.eventForm.valid) {
+      Object.assign(this.event, this.eventForm.value)
+      console.log(this.event)
```
KrisJordan, line 103: Looks like we've got some errant `console.log`s

frontend/src/app/event/event-editor/event-editor.component.ts
```diff
@@ -0,0 +1,141 @@
...
+      if (this.event.id == null) {
+        console.log("Event Updated");
```
KrisJordan, line 105: Another one :)

frontend/src/app/event/event.resolver.ts
```diff
@@ -0,0 +1,35 @@
...
+import { EventService } from "./event.service";
+
+/** This resolver injects the list of events into the events component. */
+export const eventResolver: ResolveFn<Event[] | undefined> = (route, state) => {
+    return inject(EventService).getEvents();
+}
+
+/** This resolver injects an event into the events detail component. */
+export const eventDetailResolver: ResolveFn<Event | undefined> = (route, state) => {
+    if(route.paramMap.get("id") != "-1") {
```
KrisJordan, line 22: Is routing the motivation for -1? I would maybe choose `new` as the default new form route, rather than `-1`, if it doesn't create too much complication.

frontend/src/app/event/widgets/event-detail-card/event-detail-card.widget.ts
```diff
@@ -0,0 +1,58 @@
...
+        return this.permission.check('organization.events.manage', `organization/${this.event.organization_id!}`);
+    }
+
+    /** Handler for when the share button is pressed
+     *  This function copies the permalink to the event to the user's
+     *  clipboard.
+     */
+    onShareButtonClick() {
+        // Write the URL to the clipboard
+        navigator.clipboard.writeText("https://csxl.unc.edu/events/" + this.event.id);
```
KrisJordan, line 40: Use `window.location.host` here rather than hard-coding `csxl.unc.edu`

frontend/src/app/organization/organization.resolver.ts
```diff
@@ -34,7 +39,8 @@ export const organizationDetailResolver: ResolveFn<Organization | undefined> = (
             linked_in: "",
             youtube: "",
             heel_life: "",
-            public: false
-          };
+            public: false,
+            events: null
```
KrisJordan, line 43: Should this be an empty array rather than null?

KrisJordan (changes requested):
This is looking really great!

There are a number of minor nits we should knock out ahead of student hand-off, if possible. Of high importance are the warnings emitted from the deprecated SqlAlchemy `get` API.

backend/api/events.py
```diff
@@ -0,0 +1,145 @@
...
+        event_service: a valid EventService
+
+    Returns:
+        list[EventDetails]: All `EventDetails`s in the `Event` database table from a specific organization
+    """
+
+    # Return all events
+    return event_service.get_events_from_organization(slug)
+
+@api.post("", response_model=EventDetails, tags=['Events'])
```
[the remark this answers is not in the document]
ajaygandecha: I believe it provides documentation on the /docs page for the shape of the response!

backend/entities/organization_entity.py
ajaygandecha: We found that we needed "all" for the cascading to work but will double check

backend/api/events.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+
+    # Try to create event
+    try:
+        # Return created event
+        return event_service.create(subject, event)
+    except Exception as e:
+        # Raise 422 exception if creation fails
+        # - This would occur if the request body is shaped incorrectly
+        raise HTTPException(status_code=422, detail=str(e))
```
ajaygandecha, line 73: @KrisJordan I am curious if this should also use a different Exception, or if we should create one thrown in `event_service` that is handled by the middleware.

backend/services/event.py
ajaygandecha: Order is actually needed here in order since we need the organization ID to run the permission check!

backend/models/event.py
ajaygandecha: This is required in order to post events! I originally created a `NewEvent` model for this purpose but it was removed later in the testing branch.

backend/entities/event_entity.py
ajaygandecha: As mentioned in a comment below, the only way this can work is to have a `NewEvent` model without `organization_id` and then have `EventDetails` inherit from that. Do you think this is what we should move towards?

backend/entities/organization_entity.py
ajaygandecha: @KrisJordan this is a good point. We used this convention to prevent using two API calls on the organization details page.

frontend/src/app/organization/organization.resolver.ts
ajaygandecha: @KrisJordan I think by leaving it null, it matches the shape of the `Event` backend model. May be wrong though

backend/api/events.py
KrisJordan: Yes, ideally it should. The docs also do not reflect the Raises section of the docstring.

backend/api/events.py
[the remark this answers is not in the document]
KrisJordan: Nice; we should sprinkle these around in other routes where they're missing. This is not something I've been doing.

backend/api/events.py
ajaygandecha: @KrisJordan would something like InvalidEventException, or potentially more general like InvalidRequestException be okay?

backend/entities/event_entity.py
KrisJordan: My argument would be to have `Event` not have the relation/id and then let the relation/id reside in `EventDetails#organization.id`. Effectively, anywhere you actually need do anything in the model with the related org, you would have something like:

~~~
{ ...
  organization: {id: 1, ...}
}
~~~

In the modelling, this could be an OrganizationPartial (removing required fields so you only had to worry about 1.

The high-level point being this is an example of where we want the model abstraction layer to hide the detail of the join column identity.

backend/entities/organization_entity.py
KrisJordan: This is fine for the time being, but a feature extension we should prioritize next sprint is resolving pagination of events (esp backwards looking) and improving this endpoint so that it only gives forward looking events.

backend/entities/event_entity.py
ajaygandecha: @KrisJordan in the create event service function, would there be another way we would want to check for permissions then? We can't necessarily pass EventDetails as the request model because it would lack an Organization, and currently we test for permissions using the `organization_id` passed in through `Events`.

ajaygandecha: (Specifically referencing line 57 in `/services/event.py`

KrisJordan (approved):
Looking good to me, there are a few comments that we should circle back on in a future PR.

Primarily: 

1. Thinking about the data modeling of event vs. event details
2. Double checking the need for `children` attribute in `EventDetailsComponent`
3. Pagination of past events (and only showing upcoming events) on org details / events page

PR #115, opened 2023-10-22
Install ESLint and Prettier

atoneyd:
This PR installs ESLint and Prettier to enforce consistent frontend styling, warn you about bugs, and fix style errors. I chose these specifically because they are among the most commonly used in web development.

## Major Changes
- This PR only installs these two tools, and does not run them.
- Run the command `ng lint` to view current linting errors and `ng lint --fix` to automatically fix those errors. 
- The fix command should be run at a later point to avoid merge conflicts as it will modify many files to fit the styling constraints.

## Testing
Ran `ng lint` to confirm that style errors were caught. Previously ran `ng lint --fix` to confirm that these errors could be automatically fixed by the tool, and then discarded those changes to prevent conflicts.

## Notes:
The formatting rules enforced by prettier can be found under the file `.prettierrc.json`. Rules chosen include a tab width of 4, consistent use of single quotes, a printWidth of 80, etc. I found these to be standard rules used widely across angular projects. In the JavaScript world a tab width of 2 seems to be more common but I changed it to 4 as that seems to be what most of our files are currently using. However, I would propose changing that to 2 if others are fine with it.


ajaygandecha (approved):
LGTM - ready to merge. Just a note, we should not run this linter until after the Event Editor PR has been merged and once all merge conflicts have been resolved.

ajaygandecha:
Pinging @KrisJordan to ensure that he approves this particular plugin --

PR #116, opened 2023-10-22
Event Management Functionality

jadekeegan:
This major PR provides functionality for creating, editing, and deleting events.

## Major Changes
- Added the Event Editor, which allows users to create or update an event.
- Added a list of Events to the Organization Details page to allow users to view an organization's events.
- Added an "Add Event" button to the list of Events on the Organization Details Page.
- Added "Edit Event" and "Delete Event" buttons to the Event Details Cards.
- Implemented appropriate Admin permissions to ensure that only users with appropriate access can view event management pages/buttons.
- Changed the backend event permissions to use `organization.events.manage` for all post/put/delete methods to improve usage in the frontend when granting event management permissions.

## Testing
All changes were tested locally by running `honcho start`. 

First, create a new permission for the Ambassador role with the following information: `organization.events.manage`, `organizations/[organization ID of your choice]`. To test event creation, go to the Organization Details page and click the Add icon. To test event editing and deletion, go to the Events Page (on desktop) or go to an Event Details Page (mobile or desktop). Then, click the Edit or Delete icon to test. Be sure to test as both an Ambassador user and a regular user to ensure that the permissions are functional!

## Developer Notes
An RXJS object for Event will need to be created in the future to ensure that deleting events when you are on the Events page properly reloads the Event list. Currently, we use a manual reload.

Because routes are currently mounted based on which module they are part of, the route for the event editor is very messy. Currently, it is `events/organizations/:organization_slug/events/:events_id/edit`. However, we need the organization slug to associate the event to an organization, so that path parameter needs to stay. The ideal hierarchy should be `organizations/:organization_slug/events/:events_id/edit`.


frontend/src/app/event/widgets/event-detail-card/event-detail-card.widget.ts
```diff
@@ -34,4 +41,18 @@ export class EventDetailCard {
...
+    /** Delete the given event object using the Event Service's deleteEvent method
+     * @param event: Event representing the updated event
+     * @returns void
+     */
+    deleteEvent(event: Event): void {
+        let confirmDelete = this.snackBar.open("Are you sure you want to delete this event?", "Delete");
+        confirmDelete.onAction().subscribe(() => {
+            this.eventService.deleteEvent(event).subscribe(() => {
+            this.snackBar.open("Event Deleted", "", { duration: 2000 })
+            location.reload();
```
ajaygandecha, line 54: Not for this PR and already discussed offline - but this is for @KrisJordan in case he sees, we will attempt to implement an `RxEvent` object to remove this extra page reload. We decided that it would be good not to block all other further progress for this one optimization.

frontend/src/app/organization/organization-details/organization-details.component.ts
```diff
@@ -41,11 +45,18 @@ export class OrganizationDetailsComponent {
   /** The organization to show */
   public organization: Organization;
 
+  /** Store a map of days to a list of events for that day */
+  public eventsPerDay: [string, Event[]][];
+
+  public eventCreationPermission$: Observable<boolean>;
```
ajaygandecha, line 51: Documentation missing on this property!

frontend/src/app/shared/event-list/event-list.widget.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    /** Whether or not to disable the links on the page */
+    @Input() disableLinks: boolean = false;
+
+    @Input() showHeader: boolean = false;
+
+    /** Whether or not to disable the event creation button */
+    @Input() showCreateButton: boolean = false;
+
+    @Output() onCardClicked: EventEmitter<Event> = new EventEmitter()
```
ajaygandecha, line 41: Documentation also needed for this.. think this may be my fault. You can just say:
```ts
/** Event binding for the card's on click action */
```

frontend/src/app/shared/event-list/event-list.widget.css
ajaygandecha, line 1: Slight issue here that has cascaded to the events view. I will fix in a commit.

frontend/src/app/event/event-editor/event-editor.component.ts
```diff
@@ -0,0 +1,140 @@
...
+    this.adminPermission$ = this.permission.check('organization.events.manage', `organization/${this.organization!.id}`);
+  }
+
+  /** Event handler to handle submitting the Create Event Form.
+   * @returns {void}
+  */
+  onSubmit = () => {
+    if (this.eventForm.valid) {
+      Object.assign(this.event, this.eventForm.value)
+      if (this.event.id == -1) {
```
ajaygandecha, line 103: This comment is based on the error I wrote about in the review. Here, `this.event` actually is being initialized to `null` by the `Object.assign(...)` call rather than `-1`. Changing either the conditional or the line above should solve this problem!

As a result, I think this condition always returning `false` is what was calling the `PUT` (update) API call rather than the `POST` (create) one.

ajaygandecha (changes requested):
Great work @jadekeegan! Only two comments were left for documentation and I will add that in a commit immediately following this review.

The code looks good to me, but I am actually getting a 404 error when attempting to add an event to CADS when testing in `honcho start`. Specifically, it appears during the `PUT /api/events` call. I have left a comment pinpoint where specifically I think this problem is occurring!

ajaygandecha:
Also, uh oh... Seeing `ExpressionChangedAfterItHasBeenCheckedError` coming up again.. Forgot to comment that in the original review. Seems to be happening on the Organization page, curiously. Something is changing from `true` --> `false`.

ajaygandecha (approved):
LGTM now

briannata (approved):
LGTM!

PR #119, opened 2023-10-24
Add ability to create roles as an Admin

KrisJordan:
Add the ability for administrators to create new roles in the backend User Admin UI.

A new button is added to the Roles Table:

image

We use system prompt widget to gather the name of the new role. Upon providing a name, the new role is created and the user is redirected to the role editor page.


jadekeegan (approved):
LGTM! Ready to merge!

PR #120, opened 2023-10-24
Fix Frontend Visual Bugs in Production

ajaygandecha:
This PR fixes two visual bugs in the latest release.

## Changes:
- Fixed font size of the Organization Detail page from increasing when on a wide enough monitor.
- Add a short sentence to indicate when there are no upcoming events.


KrisJordan (commented):
LGTM

PR #121, opened 2023-10-25
Fix Sorting Bug for Events Page

ajaygandecha:
This PR fixes a major bug that prevented the events page from sorting correctly.

## Major Changes
- Reimplement date sorting into the `EventFilterPipe`.
- Create the `EventJson` model to represent the exact types of data retrieved from the API.
    This was needed because dates from the API are returned as strings and are not implicitly coerced
    into the `Date` type.
- Create the `parseEventJson` pipeable function to convert `EventJson` models to `Event` models.


KrisJordan (approved):
LGTM

PR #123, opened 2023-10-25
Fix Organization Detail Page Crash

ajaygandecha:
This PR fixes a critical bug preventing the Organization Detail page from loading.

Specifically, it creates a new resolver for event data to receive that directly from the API rather than through the Organization object.


KrisJordan (commented):
LGTM

PR #128, opened 2023-10-25
Migrate CSXL Application to Cohesive Design

ajaygandecha:
This PR changes the frontend Angular Material styles of the application.

## Major Changes
- Migrated all Material Cards to use the outline style.
- Migrated all Material Buttons to use the stroked style.
- Migrated all Material Input Fields to use the outline style.

## Outstanding
- The Coworking Operating Hours widget uses an expansion panel - this does not have the ability to change to a different style! @KrisJordan, any thoughts on potentially using a different Angular element here?

### DIFF
--- a/frontend/src/app/about/about.component.html
+++ b/frontend/src/app/about/about.component.html
@@ -1,4 +1,4 @@
-<mat-card class="content">
+<mat-card class="content" appearance="outlined">
   <mat-card-content>
     <mat-card-subtitle>Find us in Sitterson 156!</mat-card-subtitle>
     <p>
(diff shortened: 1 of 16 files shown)


KrisJordan (approved):
LGTM

KrisJordan:
The Operating Hours Widget in Coworking was really meant to be a placeholder and temporary (it's ugly). We pretty quickly will need the ability to add new hours to the database (by Monday lol). If this is something we want to sprint on real quick, I'd be up for it. We don't necessarily need the solution for adding hours by Monday (I can do it in a script or by connecting to the database directly), but having a design that's ready for hours to be able to be added by an admin/ambassador from the widget would be great.

PR #129, opened 2023-10-25
Add Right Padding to Org Details Event List

ajaygandecha:
This commit adds right padding to the Organization Details Event List to make padding on mobile devices a bit more even.


jadekeegan (approved):
LGTM!!

jadekeegan:
Not related to this PR specifically, but we should move around the buttons/links.
image

The current position looks a little strange.

At the very least, maybe we could move the edit button to the upper right corner of the card?

ajaygandecha:
@jadekeegan that can be done if everyone agrees !

PR #151, opened 2023-11-08
Improve UX When Navigating to an Invalid Organization

### LINKED ISSUE #83: Improve UX When Navigating to an Invalid Organization

ajaygandecha:
This PR aims to improve the UX when a user navigates to an invalid organization. 

## Major Changes
- Rework the organization resolver to return `undefined` upon an error retrieving an organization, allowing us to handle this scenario in the frontend without immediately causing a crash. This uses RxJS's `pipe`, `catchError`, and `of` functions.
- Add a new widget called `OrganizationNotFoundCard` that shows only when an organization is not found.

## Testing
- This feature was tested on the frontend using `honcho start`.

*This PR resolves #83*

### DIFF
--- a/frontend/src/app/organization/organization-details/organization-details.component.html
+++ b/frontend/src/app/organization/organization-details/organization-details.component.html
@@ -8,7 +8,7 @@
 </div>
 </div> -->
 
-<div>
+<div *ngIf="organization != undefined">
   <organization-details-info-card
     [organization]="organization"
     [profile]="profile" />
--- a/frontend/src/app/organization/organization-details/organization-details.component.html
+++ b/frontend/src/app/organization/organization-details/organization-details.component.html
@@ -8,9 +8,9 @@
 </div>
 </div> -->
 
-<organization-not-found-card *ngIf="organization == undefined"/>
+<organization-not-found-card *ngIf="organization == undefined; else eventDetails"/>
 
-<div *ngIf="organization != undefined">
+<ng-template #eventDetails>
   <organization-details-info-card
     [organization]="organization"
     [profile]="profile" />
@@ -24,4 +24,4 @@
       [showHeader]="true"
       [showCreateButton]="(eventCreationPermission$ | async)!" />
   </div>
-</div>
+</ng-template>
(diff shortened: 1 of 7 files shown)


frontend/src/app/organization/organization-details/organization-details.component.html
```diff
@@ -8,7 +8,9 @@
 </div>
 </div> -->
 
-<div>
+<organization-not-found-card *ngIf="organization == undefined"/>
+
+<div *ngIf="organization != undefined">
```
KrisJordan, line 13: This would be more idiomatic as an else to the above ngIf

frontend/src/app/organization/organization.resolver.ts
```diff
@@ -48,8 +44,17 @@ export const organizationDetailResolver: ResolveFn<Organization | undefined> = (
...
       events: null
-    };
+    }
   }
+
+  // Otherwise, return the organization.
+  // If there is an error, return undefined
+  return inject(OrganizationService).getOrganization(route.paramMap.get('slug')!)
+      .pipe(catchError((error) => {
+        console.log(error)
```
KrisJordan, line 54: Should this be removed?

KrisJordan (changes requested):
The ngIf/else change should be made, but it's minor. This is looking great!

frontend/src/app/organization/organization.resolver.ts
ajaygandecha: I actually added this `console.log` to this so that there is some diagnostic showing the error that occurs! Otherwise, pretty sure it just returns undefined - but, I can remove this or improve it if it does not follow best practices

KrisJordan: We need to identify what the best, app-wide practice is for surfacing errors easily. Fine for now.

KrisJordan (approved):
LGTM

PR #154, opened 2023-11-09
[DRAFT] [DO NOT MERGE] Implement Audit Log Feature to the CSXL Application - Backend

ajaygandecha:
This PR introduces backend entities, models, services, and tests for implementing the audit log feature. The audit log would be an admin-level feature that tracks the usage of major mutating API calls (for example, editing organizations, deleting events, etc). This allows the administrators of the CSXL Application to have a history of these calls in case a user either maliciously uses their permissions or if unauthorized users gain access to admin or ambassador level permissions. Note, at the moment, logs are not intended for `GET` API calls, however we can easily implement this for features such as retrieving sensitive user information.

## Major Changes
- Implemented the `LogEntity` in the backend to represent the new `log` table in the database. This entity has a one-to-many relationship with the `User` table, where each log is related to a single user, but a user can have many logs.
    - The `log` table stores a primary key `id`, the `user_id` representing the user attached to the log, and a short `description` field to describe the changes that were made.
    - We can consider changing this to specific API routes if needed, but I thought that a description would be more readable.
    - We can also consider passing request bodies here, although I thought that this may not be the most efficient use of space. Thoughts @KrisJordan?
- Implemented a `Log` and `LogDetail` model to represent the data for FastAPI.
- Implemented the `LogService` to control reading and adding logs to the database.
- Added test data and unit tests for the logs.

## Testing
- Tested the service using `pytest` (100% passing)

## Still To-Do
- Add calls to the `LogService` to each mutating service method for other features. Wanted to confirm that this is the best approach @KrisJordan! Probably does not make sense to expose any other APIs for logs except `GET`.
- Create a `GET` API for the log service.

### DIFF
--- a/backend/entities/log_entity.py
+++ b/backend/entities/log_entity.py
@@ -3,6 +3,8 @@
 from sqlalchemy import Integer, String, ForeignKey
 from sqlalchemy.orm import Mapped, mapped_column, relationship
 from .entity_base import EntityBase
+from typing import Self
+from ..models.log import Log, LogDetails
 
 __authors__ = ["Ajay Gandecha"]
 __copyright__ = "Copyright 2023"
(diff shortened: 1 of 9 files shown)


KrisJordan:
This was headed in a great direction and is a promising start, but after some discussion we resolved to investigate whether we can implement application-wide logging that uses CloudApps' logging infrastructure (logstash/kibana) to solve our immediate needs. If we learn that it is not a viable path, lets reopen this and continue.

PR #155, opened 2023-11-10
Pass Initial Selected Event Through Pipe

ajaygandecha:
This PR fixes a bug where the initially selected event would occasionally show a past event rather than the most upcoming event.


KrisJordan (approved):
Lgtm

PR #157, opened 2023-11-10
Fix Organization Permissions

jadekeegan:
This pull request updates the frontend and backend to improve the organization feature permissions and make them specific to create/update/delete actions.

## Major Changes
- Updated the backend Organization Service to have more specific action/resource strings for permission checks on the post/put/delete API routes (following the admin permissions doc).
- Updated the Organization Details Card widget to use the accurate update permission check on the "Edit" button.
- Updated the Organization Editor to implement a `CanActivateFn` to determine whether or not the organization editor is viewable for the registered user.
- Update the permissions doc to reflect the use of `organization_slug` instead of `organization_id` in the resource string for `organization.update`.
- Improve the Organization Editor to have different titles/confirmation messages based on whether the user is creating or deleting an event.

## Testing
These changes have been thoroughly tested on the local server using `honcho start`. To replicate these tests:
1. Start the local server.
2. As the Root user, go to the Admin Organization tab.
3. Create and delete an organization, and ensure that it works properly.
4. Add a role on the Admin Roles page corresponding to one of the organizations.
5. Grant the following permission: `organization.update`, `organization/{organization_slug}`, where the slug corresponds to the organization you just created a role for.
6. Assign a user to the role, and then log in as that user.
7. Find the organization you granted permissions to and click on the details for it. Check to ensure that an edit button is there and functional.
8. Go to the `/docs` page and ensure that the behavior is correct as a Root user, a user with specific organization permissions, and a user with no permissions.

## Developer Notes
Currently, navigating to the `organizations/new/edit` path without the appropriate permissions just displays a blank page rather than routing to a separate page. This is not a major issue as this edge case is very rare, but it should be noted.

The organization permissions were also changed to use the organization_slug in the resource string rather than organization_id. This is because the canActivateFn only has access to the route parameters, which in our case is just the organization_slug. This change works, but it does make the permissions a little inconsistent with the event management permissions.


atoneyd (changes requested):
Looking good. Pytests are currently breaking since we've removed the OrganizationNotFoundException, so those tests also need to be refactored before merge.

Regarding your note about the blank page when you navigate to organizations/new/edit, I think we should handle this similarly to what is done with the organization not found card. I was thinking it would be good for us to have a Page Not Found card with a 'go back' button that we can use in general, which could just replace the org not found card so that we can use it anywhere(I feel like this is what most websites do). We can discuss this further though/push this issue to another ticket.

atoneyd (approved):
LGTM, letting @KrisJordan give final approval

KrisJordan (approved):
LGTNM

PR #161, opened 2023-11-11
Implement Event Registration Backend

ajaygandecha:
This major PR provides the implementation of the Python backend for the Events Registration feature. This includes the SQLAlchemy Entities and Pydantic models, backend services, FastAPI routes, and test data. 

This PR will ultimately serve to port over the ability for users to register for events, improving on the initial attempt from the A3 final project last semester.

- Added the `EventRegistration` SQLAlchemy Entity to model how registrations should be stored in the PostgreSQL database.
- Established a many-to-many relationship between the `user` and `events` tables via the secondary join association table `event-registrations`.
- Added the `EventRegistration` and `NewEventRegistration` Pydantic models to represent the shape of events data to be exposed by the APIs. The `NewEventRegistration ` model exists only for the creation of registrations by the component.
### Scripts Changes
- Added test data for event registrations. This data should populate when the `test_reset.py` script runs.
### API Changes
Exposes the following APIs to the Angular Frontend via FastAPI:
- POST `api/events/register/:id`: Adds a new event registration to the database.
- DELETE `api/events/register/:id`: Deletes an event registration to the database.

*NOTE: registrations are accessible through the GET APIs for users and for events.*

### Service Changes
- Added the `register()` and `unregister()` functions to the `EventService` to manage event registrations.

## Testing
These changes were thoroughly tested. New PyTests were added to thoroughly check the backend functionality for event registrations works. Test pass 100%.

## Future Considerations
- Frontend will be added in a later PR.
- We can consider adding more fields and details such as timestamps to event registrations.
- @KrisJordan, I think this can serve as a model for COMP 423 students to implement many-to-many relationships in their projects. We may be able to mention this on Monday and encourage them to pull from `upstream/main` if this looks good enough to serve as a model.


KrisJordan:
A few important changes we'll want to make to the design of this API and implementation:

1. We need to remove registrations from `EventDetails`. The reasoning is it leaks user data to all users (including unregistered) who are listing events.
2. We need to add an endpoint for getting the list of registrations for an event (guarded by permissions) since we're removing it from EventDetails.
3. We should roll back the changes to the UserAPI (including EventRegistrations) and establish an API endpoint for getting a user's registrations (is ideally going to have a similar pagination/history API as what top-level events will have, but if we want to launch with listing all, that's fine)
4. I'd like to try refactoring the EventRegistration models to avoid using `event_id` and `user_id`. These feel like they should be hidden behind the abstraction and are entity-level concerns. I can take a stab at this.

KrisJordan:
Going to take a stab at #1 and #2 next.

backend/services/event.py
```diff
@@ -199,3 +220,142 @@ def delete(self, subject: User, id: int) -> None:
...
+                "user.event_registrations",
+                f"user/{user.id}",
+            )
+
+        registration_entities = (
+            self._session.query(EventRegistrationEntity)
+            .where(EventRegistrationEntity.user_id == user.id)
+            .join(EventEntity, EventRegistrationEntity.event_id == EventEntity.id)
+            .where(EventEntity.time >= time_range.start)
+            .where(EventEntity.time < time_range.end)
```
ajaygandecha, line 358: I wonder if there is a way to make some sort of helper or query builder in Python that would allow us to more easily compare date ranges! Unsure if this exists though.

backend/services/user.py
```diff
@@ -49,6 +50,24 @@ def get(self, pid: int) -> UserDetails | None:
...
+        Args:
+            id: The ID of the user.
+
+        Returns:
+            User
+
+        Raises:
+            ResourceNotFoundException if the User ID is not found
+        """
+        user_entity = self._session.get(UserEntity, id)
```
ajaygandecha, line 65: Do we need permissions for this particular call? Such as if there is sensitive data here

ajaygandecha (commented):
Looks great to me @KrisJordan! Just had a few comments but nothing blocking from what I can see.

backend/services/user.py
KrisJordan: Hmm interesting question. In the usage, we're not leaking anything, but this point is well taken.

I think `UserService` is due for a cleanup and refactor (such as taking up the idiom of using `subject` for most methods).

The use case we care about here is an organizer or admin being able to add an attendee. For this reason, and some other scenarios (such as searching for friends), I believe we'll want registered users to be able to lookup other users (but perhaps only get a subset of their profile data).

This is something we should come back to and tidy up after the semester.

backend/services/event.py
KrisJordan: This is worth looking into, certainly we could write a helper ourself that has a form of something like:

`within_time_range(query, EventEntity.time)`

KrisJordan:
From #164, the scope of this backend API is starting to expand as some additional thought is put into use cases we need. Rather than merging into `main` before it's stable, and without a frontend, we'll demonstrate some concepts from this PR branch in lecture.

Some of the key pieces we need to resolve ahead of merging to `main`:

* Flag for an Event determining whether users can register for it.
* Max capacity of registrations for an Event.
* Surfacing of registration count / seats available in API (and possibly denormalized count state in Event?) - better: advanced SQLAlchemy query that joins on an aggregate function. Count should not include event organizer registrations.
* Logic for not allowing registrations beyond registration limit.
* Logic for establishing who event organizers are (does the subject who creates an event automatically get listed as an organizer)?

KrisJordan:
One other nit, is the data modeling of Event correct with `id: int | None`. I know I did this in `User`, but I'm thinking this was a mistake in modeling. It leads to many places where we have to check for None, use the `|` bar, or suppress errors (such as in testing file).

I tend to wonder if we have something like an `UnsavedEvent` that has everything except an `id`, then `Event` extends with an `id`.

ajaygandecha:
@KrisJordan re-requesting review! I have made an initial attempt for all of the bulletpoints mentioned and wrote tests for them (all pass atm)

backend/models/event_registration.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Pydantic model to represent an `EventRegistration`.
+
+    This model is based on the `EventRegistrationEntity` model, which
+    defines the shape of the `EventRegistration` table in the PostgreSQL database
+    """
+
+    event: Event | None
+    user: User | None
+    is_organizer: bool = False
```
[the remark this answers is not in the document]
ajaygandecha: Agree, I think this can be safely removed!

ajaygandecha:
@KrisJordan any updates on this, are we good to merge? I can also create a branch called `feature/event-registration` off of main as an intermediate step if needed here. Unsure if any migrations will be needed.

backend/models/event_registration.py
```diff
@@ -0,0 +1,39 @@
...
+    defines the shape of the `EventRegistration` table in the PostgreSQL database
+    """
+
+    event: Event
+    user: User
+    is_organizer: bool = False
+
+
+class EventRegistrationStatus(BaseModel):
+    registration_count: int
```
KrisJordan, line 39: Should this actually be an extension of the `Event` model, like `EventDetails` or `EventWithRegistrationCount`? It feels a little strange to have it be just an integer (if used as just an integer, the return type from an API call could probably be just an integer rather than a single-field object).

KrisJordan (approved):
LGTM with one minor nit/question regarding the registration count model...

ajaygandecha:
@KrisJordan, I am closing this ticket since there have been a lot of event registration backend changes on #212 !

PR #162, opened 2023-11-13
Modifications to the event registration feature PR

KrisJordan:
This PR is a WIP that targets changes to the Event Registration Backend features.

### DIFF
--- a/backend/api/events.py
+++ b/backend/api/events.py
@@ -136,7 +136,7 @@ def delete_event(
     event_service.delete(subject, id)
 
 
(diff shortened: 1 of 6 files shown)


KrisJordan:
@ajaygandecha I'm iterating on your original design here and think this is good to the point of being able to merge into your branch. There are a few more things I think we'll want to do before deploying, but I'll bring that conversation back to the feature branch.

KrisJordan:
For the purposes of expediency and working toward shipping ahead of class tomorrow, I'm going to go ahead and merge this into the other branch and keep chipping away!

PR #163, opened 2023-11-14
Refactor user registrations to live separate from UserDetails

KrisJordan:
I'm moving forward with refactoring how to approach UserDetails / registrations @ajaygandecha


KrisJordan:
Hmm, I did something really suspect here...

PR #164, opened 2023-11-14
Protect EventRegistrations attendees from accidental API disclosure

KrisJordan:
We need registrations to events to be limited to users with permission to view. This PR is not ready for merge, but is the start of progress on this front.

### DIFF
--- a/backend/api/events.py
+++ b/backend/api/events.py
@@ -234,6 +234,26 @@ def get_registration_status(
         return event_registration
 
 
+@api.get("/{event_id}/registrations", tags=["Events"])
+def get_event_registrations_status(
+    event_id: int,
+    subject: User = Depends(registered_user),
+    event_service: EventService = Depends(),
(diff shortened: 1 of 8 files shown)


KrisJordan:
Making progress on this improvement. Added the notion of a registration being as an `organizer` which we can use in some of the feature-specific authorization logic. It does reveal some new concerns, though, such as being sure a registration can't come in and make itself an organizer (unless the registration is being performed on behalf of someone else). Alternatively, maybe the API is to elevate a regular registration to an organizer, or back again, as long as the right authorization rules are enforced.

Ultimately, I don't think this is going to be mergable into stage in time for students and will need to serve as some examples we can link to in this repository as WIP. Some additional logical concerns we'll need to address at the API level include:

- A flag in Event to make it registerable
- A cap on the number of registrations for an event
- Reporting on spots available / open (not counting organizers...)

PR #165, opened 2023-11-15
Create Documentation for SQLAlchemy Database Relationships for COMP 423

ajaygandecha:
This PR creates a reading that explains database relationships and model circularity in the context of the CSXL web application. The purpose of this reading is to help students gain a better understanding of SQLAlchemy and SQLAlchemy relationships, as well as working with entities and models in their own final projects.


KrisJordan:
This is looking great! I will take a closer read and perhaps make some edits shortly.

One thing I'll point out which may be worth updating is establishing the primary key of the `EventRegistrationEntity` to be a compound key of the two related keys. This is feasible because we do not permit duplicates (e.g. multiple registrations of the same user for the same event) so those pairs are unique and can serve as a PK. This is true of _many_ many-to-many relationships, but certainly not all (e.g. if you were tracking certain activities like "(user) went for (run)" there would be many run associations, assuming that's in an activity type table or something, so you'd need a unique primary key).

Here's what the updated Entity design looks like: https://github.com/unc-csxl/csxl.unc.edu/pull/161/files#diff-e243578aa359f08d4aba5bcffdd3967a6c595677f533812c8a7bb6a54f0c3fb7R30-R38

ajaygandecha:
No problem! Also @KrisJordan a question I had about the compound primary key structure before writing this reading, how would you query from the association table using `.get(EventRegistrationEntity, id)`?

Do not mind updating the reading to include this! I was considering using this because that was the example we will be merging in, just wanted to make sure I do it in a way that was less confusing (not sure if people will get lost with the primary and foreign keys being on the same field - but we can try)!

KrisJordan:
Sorry for the delay, just wanted to test and confirm:

`session.get(EventRegistrationEntity, (1, 2))`

KrisJordan:
Is it intentional that the docs deviate from the `UserEntity` with respect to the `primary_key` being `pid` as opposed to `id` in the code base? For demo purposes, I don't think a big deal, but if it's accidental, might be a quick fix.

ajaygandecha:
@KrisJordan it was intentional mainly for instructional purposes! I can change it if needed though.

KrisJordan:
No, that sounds good to me, was just checking!

PR #178, opened 2023-11-18
Create Documentation about Angular Resolvers for COMP 423

ajaygandecha:
This PR creates an additional documentation reading that explains Angular Resolvers to students in COMP 423.


KrisJordan (approved):
LGTM

PR #179, opened 2023-11-21
Create Documentation about Middleware Error Handling for COMP 423

ajaygandecha:
This PR creates an additional documentation reading that explains error handling using the middleware to students in COMP 423.

The page is accessible [here](https://github.com/unc-csxl/csxl.unc.edu/blob/docs/middleware/docs/backend/middleware.md).


KrisJordan (approved):
This is a great overview covering the motivation for why this is useful and the implications / usage of middleware. Great work!

PR #180, opened 2023-11-22
Create Documentation about RxObject for COMP 423

ajaygandecha:
This PR creates an additional documentation reading that explains how to use RxObject to students in COMP 423.

The page is accessible [here](https://github.com/unc-csxl/csxl.unc.edu/blob/docs/rxobject/docs/angular/rxobject.md).


docs/angular/rxobject.md
```diff
@@ -0,0 +1,416 @@
...
+let subject: Subject<string> = new ReplaySubject(1);
+let value$: Observable<string> = subject.asObservable();
+
+// Access latest data from the subject
+value$.subscribe((stringValue) => {
+  console.log(stringValue);
+})
+```
+
+As you can see, the code above creates an `Observable` from the `ReplaySubject` which we can subscribe to!
```
KrisJordan, line 129: This is where it can be pointed out the key distinction -- new values can be pushed to `subject`, but `value$` is read-only.

docs/angular/rxobject.md
```diff
@@ -0,0 +1,416 @@
...
+
+```ts
+let subject: Subject<string> = new ReplaySubject(1);
+```
+
+> **NOTE:** The subject created above stores *strings* as its value. However, you can replace this data type with any data type you like.
+
+It is important to note that the `ReplaySubject()` constructor takes a *parameter*. This parameter is the ***buffer size***, or the amount of past data that it should store! In this case, we only want to keep the latest (most up-to-date) data in our subject, so we can keep our buffer size as `1`.
+
+Once we created the `ReplaySubject` object, we can then *feed data into it* (set its value) using the `.next()` method. This method also ***notifies*** all observers pointing to the subject.
```
KrisJordan, line 108: This is what I was getting at above, but I wonder if this is an important enough distinction of `Subject` to be elevated earlier up as two key benefits/differences of `Subject` vs `Observable` -- 1) ability to push/feed data into the subject and 2) data muticast back out of it to observers

docs/angular/rxobject.md
```diff
@@ -0,0 +1,416 @@
...
+
+Notice that we also make this class **`abstract`**. The core motivation behind in the `RxObject<T>` class is that we *do not instantiate it directly*. Instead, we want to have *subclasses* ***extend*** this base class for related usecases. For example, if we wanted to dynamically store data for organizations, we may want to create a subclass `RxOrganizationsList` that *extends* `RxObject<Organization[]>`. We will utilize this in the future, but for now, we are just working on the conceptual `abstract` class.
+
+From here, we know we need to successfully implement the Observer design pattern. We know that the Angular Components are going to be the *observers*, which means the only thing missing here is the *subject*. We want our `RxObject<T>` class to contain our subject.
+
+You can recall that, after examining the various options for our RxJS subject object, we decided that `ReplaySubject<T>` is the best for our use case. This is because the `ReplaySubject<T>` re-emits the latest value that is has to *new subscribers*. We will use a `ReplaySubject<T>` object to serve as the main *subject* to satisfy the requirements of the Observer design pattern.
```
KrisJordan, line 151: A key idea here is the pattern of: if we have loaded a piece of data before, let's serve that up from cache and then behind the scenes we can fire off a request for updated data and when it comes in we can emit the newer data if there are any changes.

docs/angular/rxobject.md
```diff
@@ -0,0 +1,416 @@
...
+  * been set.
+  */
+  private subject: Subject<T> = new ReplaySubject(1);
+
+}
+```
+
+As you can see, we *initialize* our `subject` field using a new `ReplaySubject` with a buffer size of `1` so that it stores just the most up-to-date value.
+
+So, if this field is private, then how do we want our Components to interact with our object? We want the Components to interact with the object by *subscribing to a simple `Observable<T>`*. The motivation behind this is so that `subject` manages the data stream and ensures that everything is working correctly behind the scenes, and that our Components do not need to worry about this concern and can work with the more general `Observable` type everywhere else. In addition, it enables the use of the `| async` pipe in HTML. Since we want our Components to access this field, we can make it public. Let's add the observable, made from our `subject`, into our object:
```
KrisJordan, line 171: Here's another place where I wonder if the notion of subject being read/write and managed internally and its observable being read-only is important. We're trying to keep the logic for notification encapsulated in the `RxObject` subclasses (through the helper method you're about to reveal).

docs/angular/rxobject.md
```diff
@@ -0,0 +1,416 @@
...
+Now, ***in our service***, we can create two new fields - one to store our `RxCoworkingObject` which allows us to dynamically update our data, and then one public observable for our components to use that is automatically updated with new data! The fields are shown below:
+
+```ts
+private status: RxCoworkingStatus = new RxCoworkingStatus();
+public status$: Observable<CoworkingStatus> = this.status.value$;
+```
+
+Notice that `status`, which is our `RxCoworkingStatus` object, is private - this is because we want to abstract all of this functionality and keep it from our Components. Instead, the only thing accessible to the components should be the *observable* that is automatically updating with new data, `status$`! Notice that this observable is simply the public `.value$` observable from our `status` object that we defined in the previous part.
+
+Finishing this functionality is easy! All we need to do now is call the `.set()` method on our `RxCoworkingStatus` object (remember, this was *inherited* from `RxObject` because `RxCoworkingStatus` extends `RxObject`). We can do this in `pollStatus()`, which is supposed to update our data! Look at the following final construction:
```
KrisJordan, line 337: Seeing these comments makes me wonder if a refactoring of RxObject is in store for simplifying things 👀 

This may be a winter break goal, but this has me wondering: _why doesn't RxObject implement `Observable<T>`_ and then we could get rid of this ugliness and just make `public status$: RxCoworkingStatus` without the private side. Now, this would mean that the public methods of `RxCoworkingStatus` are exposed outside of the service, but this might be an OK trade-off. This extra step of doing public/private _again_ feels like it's a bit overkill and would love to simplify what the service-layer implementation looks like.

As I write this out, I am also wondering if all we really want is a subclass of `Subject` that is effectively `ReplaySubject(1)` where we can directly access the last value in subclasses without having to subscribe. This is a path worth exploring at some point soon.

KrisJordan (changes requested):
This is really, really great documentation building up to our internal abstraction.

I've added some comments on suggestions-- most of which are ultimately revealed as progress is made through the documentation-- but in my first-read through it felt like some of the content could be elevated for preciseness and/or withheld until the right moment.

In reading this, it did make me question whether this is the _right_ abstraction for the job. I'm wondering if RxObject should ultimately subclass ReplaySubject, hard-code a buffersize of 1, and track/expose the last value in a field. Something to investigate over break.

ajaygandecha:
@KrisJordan, thank you for the comments here! I am going to close this PR in the meantime because it looks like a rewrite of `RxObject` is coming down the pipeline. Once we rewrite it, I will re-use and modify these docs where appropriate. Thanks for the feedback here!

PR #185, opened 2023-12-04
Add oc OpenShift Client to DevContainer

KrisJordan:
DevContainer Dockerfile now downloads and installs the oc client for OpenShift. This enables a workflow of working with staging/production within the DevContainer terminal and removes a small hurdle for students needing to download and correctly install oc. Even more optimistically this enables us to begin scripting many devops/deployment concerns in the codebase. Some thought will need to be put into guard rails to avoid accidentally scripting against the production site before scripting anything, though!

As this changes the dev Dockerfile, a rebuild of the DevContainer is needed once the commit is pulled into a dev repository.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -89,6 +89,18 @@ RUN curl https://raw.githubusercontent.com/ohmyzsh/ohmyzsh/master/tools/install.
 # Set Locale for Functional Autocompletion in zsh
 RUN sudo locale-gen en_US.UTF-8
 
+# Add `oc` OpenShift Client to DevContainer
+RUN arch="$(arch)"; \
+    case "$arch" in \
+        x86_64) export TARGET='' ;; \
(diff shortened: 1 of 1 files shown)


KrisJordan:
It'd be helpful to have CR from both x86 Mac, Apple M1+ Mac, and ideally Windows x86, as well, before merging this in. The steps to verify are switch to branch, rebuild devcontainer, then try `oc version` which should result in a `Client Version: 4. ...` printing out.

For next semester's 423, it means the hassle of downloading and installing `oc` on the host machine and all of the confusion of remembering to run `oc` commands from the host machine rather than the devcontainer will go away. Additionally, it will make it easier to begin scripting out some common steps (and setup verification checks!) on CloudApps.

ajaygandecha (approved):
This LGTM! The `oc` command works on my M1 Mac. @jadekeegan, feel free to test on Windows!

KrisJordan:
@alshayefaziz can you try verifying this on Windows?

PR #186, opened 2023-12-07
Add Generate Bearer Token feature

alshayefaziz:
This PR gives students the ability to display/copy the bearer token of a user after signing in. It removes the tedious process of having to continuously access the browsers developer tools when wanting to test the backend. I decided to add the feature to the Profiles page, because when signing in that is the page users are led to. 

Potential Improvements:
1.  Getting rid of this weird margin between the buttons and the token itself.
image

### DIFF
--- a/frontend/src/app/profile/profile-editor/profile-editor.component.html
+++ b/frontend/src/app/profile/profile-editor/profile-editor.component.html
@@ -93,24 +93,21 @@
   </ng-template>
 </mat-card>
 
-<html>
-  <mat-card appearance="outlined">
-    <mat-card-actions class="token">
-      <button mat-stroked-button type="submit" (click)="displayToken()">
-        Display Bearer Token
-      </button>
-      <button
-        class="cancel-button"
-        mat-stroked-button
-        type="submit"
-        (click)="copyToken()">
-        Copy
-      </button>
-    </mat-card-actions>
-    <div *ngIf="token">
-      <mat-card-content class="token" *ngIf="token">
-        {{ token }}</mat-card-content
-      >
-    </div>
-  </mat-card>
-</html>
+<mat-card *ngIf="profile.id" appearance="outlined">
+  <mat-card-actions class="token">
+    <button mat-stroked-button type="submit" (click)="displayToken()">
+      Display Bearer Token
+    </button>
+    <button
+      class="cancel-button"
+      mat-stroked-button
+      type="submit"
+      (click)="copyToken()">
+      Copy
+    </button>
+  </mat-card-actions>
+  <div *ngIf="showToken">
+    <mat-card-content class="token" *ngIf="token">
+      <code>{{ token }}</code></mat-card-content>
+  </div>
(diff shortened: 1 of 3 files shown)


frontend/src/app/profile/profile-editor/profile-editor.component.html
```diff
@@ -92,3 +92,25 @@
...
+      </button>
+      <button
+        class="cancel-button"
+        mat-stroked-button
+        type="submit"
+        (click)="copyToken()">
+        Copy
+      </button>
+    </mat-card-actions>
+    <div *ngIf="token">
```
KrisJordan, line 110: I may take a stab at a slight refactoring here to make this not dependent on the token such that the token can be loaded in construction.

frontend/src/app/profile/profile-editor/profile-editor.component.ts
```diff
@@ -21,6 +21,7 @@ export class ProfileEditorComponent implements OnInit {
   };
 
   public profile: Profile;
+  public token = new String();
```
KrisJordan, line 24: Thinking we can initialize this in the constructor and simplify a few things. Additional push coming for your review.

frontend/src/app/profile/profile-editor/profile-editor.component.ts
```diff
@@ -62,6 +63,15 @@ export class ProfileEditorComponent implements OnInit {
     });
   }
 
+  displayToken(): void {
+    this.token = String(localStorage.getItem('bearerToken'));
```
KrisJordan, line 67: My goal is going to be to move this to the constructor.

frontend/src/app/profile/profile-editor/profile-editor.component.ts
```diff
@@ -62,6 +63,15 @@ export class ProfileEditorComponent implements OnInit {
...
+  }
+
+  copyToken(): void {
+    navigator.clipboard.writeText(String(localStorage.getItem('bearerToken')));
+    this.snackBar.open('Token Copied', '', { duration: 2000 });
```
KrisJordan, line 72: Love seeing the snackBar opening up! We'll be able to reference the loaded token rather than reading from `localStorage` again.

KrisJordan (commented):
This is looking fantastic, I'm going to make a few quick changes in a push to get your thoughts on. Incoming.

alshayefaziz (commented):
I agree that we should have showToken in order to initialize token once in the constructor to call 'localStorage.getItem('bearerToken')'! It also allows for a much easier 'toggle' option. Lets merge!

ajaygandecha:
Hey @alshayefaziz, this looks amazing!! One very small nit is that we can have the button change from `Display Bearer Token` to `Hide Bearer Token` when it is active, but otherwise super cool and will be very helpful for students. Congrats on being the first Fall 2023 423 student to have a feature merged into the CSXL site! 🥳

PR #195, opened 2023-12-08
Create Models for Course Feature

ajaygandecha:
This major pull request makes progress on adding all of the necessary models for the courses feature to function. This includes including courses, terms (semesters), course sections, and the necessary details models to store relationships.

### DIFF
--- a/backend/entities/courses/course_entity.py
+++ b/backend/entities/courses/course_entity.py
@@ -3,6 +3,8 @@
 from sqlalchemy import Integer, String
 from sqlalchemy.orm import Mapped, mapped_column, relationship
 from ..entity_base import EntityBase
+from ...models.courses import Course
+from ...models.courses import CourseDetails
 
 __authors__ = ["Ajay Gandecha"]
 __copyright__ = "Copyright 2023"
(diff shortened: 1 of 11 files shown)


ajaygandecha:
@KrisJordan made some initial progress on the models here. The main question I had here is how we want to handle models for the join table (`user_section`). Is this something we want a model for that matches the database, or do we want a custom model here such as a SectionRosterItem or something like that?

I imagine the use case would be that we have an API that gets the roster of a course.

There are a few approaches:
- Add a new subclass of `SectionDetails` that contains information on the members of the course, and this is populated by the service function.
- Add model to represent join table data, populated with service functions.

KrisJordan:
This is where some differentiation between instructor/staff and students may come into the mix. We likely want an easy way to have limited instructor information (first/last/pronouns/email) modeled for front-end purposes. It's a little less clear how much of TA information we want in the front-end for privacy purposes: likely depends on the feature (students in the class can see, but not publicly listed). Finally, student enrollments needs to be very private by default and access control here will be important.

For now, I'd suggest focusing on the design of modeling `SectionStaff` and assuming it's generally served up as a member of `SectionDetails`. This would have a field of type of staff (instructor/uta/gta).

I think until we have a feature that needs to involve students self-associating with a class, let's leave that association unrepresentable as a model for the short term. That said, going down the road of `SectionDetailsWithRoster`, or something, where the students field is actually a list of User, _may_ make more sense than adding a special model for `Student`.

ajaygandecha:
@KrisJordan do you think it is better to keep the staff field of `SectionDetails` default to `[]` and then have the service fill the list? Without this, we would need to re-add a relationship field I think

KrisJordan:
That seems fine, yeah.

ajaygandecha:
Alright @KrisJordan, ready for initial review! 😄

KrisJordan (approved):
This lgtm!

PR #205, opened 2023-12-16
User Admin tab should Navigate to Users Table

alshayefaziz:
This PR makes it to where when a user with admin permissions clicks on User Admin, they will be directed to the Users table rather then the blank page!


KrisJordan (approved):
This lgtm! I was overthinking a future where it wasn't certain whether users would be the first tab for admins without user privileges, but we are not there yet. Merging, thanks!

PR #207, opened 2023-12-17
Implement Design for Operating Hours

ajaygandecha:
This PR implements a new design for the operating hours expansion panel on the coworking page. The new design utilized the Material `outlined` style and fits better into the style of the site. The new design also uses a Material Dialog component to display operating hours.

### Future Improvements
* We may be able to add some text specifying to the user when the XL will reopen.
* Better logic for when the XL has no future operating hours.


KrisJordan (approved):
Lgtm!

KrisJordan:
When Closed, it does not currently seem possible to see future hours to know when to expect it is open again. We should add that back in!

PR #209, opened 2023-12-17
Fix Querying Methods in Backend Service Logic

ajaygandecha:
This small PR fixes an issue in the EventService where we use two queries to retrieve events for an organization by a slug. Instead, we have moved towards using the `.join()` syntax that is now standard, making the code more concise and efficient.


KrisJordan (approved):
Nice clean up, lgtm

PR #210, opened 2023-12-17
Use get_by Convention in Organizations and Events

ajaygandecha:
As noted in the https://github.com/unc-csxl/csxl.unc.edu/pull/196 PR, @KrisJordan recommended that we standardize around the convention of calling getter backend service methods that use parameters get_by_{arg} rather than get_from_{arg}.

This PR makes this change in the backend for organizations and events.


KrisJordan (approved):
Thanks for cleaning up, lgtm

Issue #208, opened 2023-12-17
Implement the Event Registration Feature


ajaygandecha:
This ticket adds the frontend functionality to register for events based on the event backend from #160 .

PR #212, opened 2023-12-20
Implement Event Registration Feature

jadekeegan:
This PR adds buttons and functionality to the frontend to allow users to register for/unregister from events.

## Major Changes
- Added a button in the Event Detail Card for users to register/unregister for events.
- Added service methods to the Event Service for getting event registrations/registering/unregistering/etc.
- Created `EventMember` and `EventOrganizer` objects to display registered users without exposing all personal information about the users.
- Refactored `Event` models to store `is_registered`, `is_organizer`, `organizers`, and `attendees` fields. The latter two fields store `EventMember` objects to avoid user information exposure.
- Updated frontend to accommodate new backend models and API routes.
- Added an API route to retrieve a paginated list of users registered for an event.
- Added a paginated registrations card to the Event Details page to display users who are registered for an event.
- Added support on the backend and frontend for setting organizers manually (rather than just defaulting to the user creating the event).
- Surfaced the registration limit field onto the frontend and added a form control for setting/updating the registration limit. *Note: Setting the registration limit to 0 when creating/editing an event sets the can_register field to False!

### Backend
Check unit tests by running `pytest`.

### Frontend
All changes were tested thoroughly on the local server. 
1. Reset the database with `python3 -m backend.script.reset_demo` and run `honcho start`.
2. Navigate to the Events page without signing into the CSXL application. Check the Events Page, Organization Page (for CADS, HackNC, or CS+SG), and Event Details page and ensure the events appear. Ensure that registration information does not appear.
3. Log in as Sally Student with `/auth/as/sally/111111111` (or any non-Root user). Check the pages above and ensure that registration information appears. Ensure that the `Register/Unregister` buttons appropriately update based on user registration status. Be sure to stay registered for an Event!
4.  Log in as Rhonda Root with `/auth/as/root/999999999`. Check the pages above and ensure information appears. be sure to register for the same event as Sally Student on the Events page! Navigate to the Event Details page for the registered event and ensure that the registrations list appears. Ensure that users are ordered by first name (so Rhonda Root appears before Sally Student even though Rhonda Root registered after).
5. Open the Event Editor for the current event by clicking the pencil icon. Ensure that you can edit the registration limit. Ensure that you are not allowed to set to the registration limit lower than the current number of non-organizer registrations.
6. Add Sally Student as an organizer to the event and ensure it properly updates the registrations and registration count.
7. Log back in as Sally Student. Ensure you can edit the event as Sally Student. Ensure that you cannot edit the organizers of the event.

## Developer Notes
- I decided to make it so that organizers who do not have organization event management permissions (aka regular student organizers) cannot edit the organizers for an event.

## Future Considerations
- Profile page with Event Registrations coming soon once the Event Pagination PR is approved.
- The list of Registered Users for an event will be moved to a specific Event Management page with other event admin features once the Admin Gear PR is approved. Filtering and ordering for the registered users list is supported on the backend and will be surfaced on the frontend when the Event Management page is created.
- `EventOrganizer` and `EventMember` are temporary models that will likely be refactored once we discuss further how these models relate to the models in Academics.

> *This PR resolves issue #208.*


ajaygandecha:
I would like to add that this PR also contains the changes in #161 !

backend/services/event.py
```diff
@@ -189,3 +233,296 @@ def delete(self, subject: User, id: int) -> None:
...
+        events_registered_for = set(
+            [registration.event_id for registration in registrations_for_user]
+        )
+
+        return [
+            UserRegistrationStatus(
+                event_id=event.id, is_registered=event.id in events_registered_for
+            )
+            for event in event_entities
+        ]
```
KrisJordan, line 528: We will need to approach this differently. First, all events in the history of time are loaded on line 509 (and iterated over here). Second, I would argue that either this service method makes sense to return only a list of event models (perhaps without details) that a user is registered for, _or_ perhaps we have a variation on an `EventModel` where the field of `is_registered` is added to it, and there's an endpoint that a logged in user can query to get all events with registration details _as the user_. When not logged in, a different endpoint could be used for the public event listing page, and perhaps this just defaults to false for `is_registered` (or returns the more limited model). It his makes sense to talk through, happy to jump on a zoom!

backend/api/events.py
```diff
@@ -33,20 +45,20 @@ def get_events(event_service: EventService = Depends()) -> list[EventDetails]:
...
-    Parameters:
+    Args:
         slug: a valid str representing a unique Organization
         event_service: a valid EventService
 
     Returns:
         list[EventDetails]: All `EventDetails`s in the `Event` database table from a specific organization
     """
-    return event_service.get_events_by_organization(slug)
+    return event_service.get_events_from_organization(slug)
```
KrisJordan, line 61: Here's a place where I would probably refactor to two lines:

~~~
organization = organization_service.get_org_by_slug(slug)
return event_service.get_events_by_organization(organization)
~~~

It's potentially less efficient to do the extra query, however, I think it cleans up the `API` of event_service quite a bit. It's a leap to need to know the slug of an org in order to use the event service.

KrisJordan: There is another comment with more context later in the review on this, but I wonder if we should be passing in the user (or the lack of a user in a public-viewing case) and then returning models with a little more detail that include registration status.

backend/services/event.py
```diff
@@ -189,3 +233,296 @@ def delete(self, subject: User, id: int) -> None:
...
+
+        # Raise exception if event is full.
+        if event_status.registration_count >= event.registration_limit:
+            raise EventRegistrationException(event.id)
+
+        # Enable idemopotency in returning existing registration, if one exists.
+        # Permission to manage / read registration is enforced in EventService#get_registration
+        existing_registration = self.get_registration(subject, attendee, event)
+        if existing_registration:
+            return existing_registration
```
KrisJordan, line 391: Nice! Love seeing idempotent service methods 😍

backend/main.py
```diff
@@ -83,6 +87,13 @@ def resource_not_found_exception_handler(
     return JSONResponse(status_code=404, content={"message": str(e)})
 
 
+@app.exception_handler(EventRegistrationException)
+def event_registration_exception_handler(
+    request: Request, e: EventRegistrationException
+):
+    return JSONResponse(status_code=403, content={"message": str(e)})
+
+
```
KrisJordan, line 96: There's actually a more idiomatic way for feature specific exception handling to be registered! Let's keep `main.py` exceptions to those that are top-level of project.

First, add a file like this to the feature: https://github.com/unc-csxl/csxl.unc.edu/blob/main/backend/api/coworking/exception_handlers.py

Then, register the feature specific exception handler here: https://github.com/unc-csxl/csxl.unc.edu/blob/main/backend/main.py#L86-L89

backend/services/event.py
```diff
@@ -189,3 +233,296 @@ def delete(self, subject: User, id: int) -> None:
...
+            )
+
+        # Delete object and commit
+        self._session.delete(
+            self._session.get(
+                EventRegistrationEntity,
+                (event_registration.event.id, event_registration.user.id),
+            )
+        )
+        self._session.commit()
```
KrisJordan, line 435: To make this idempotent, as advertised in the docstring, how about if `event_registration` is None we just return early rather than raise?

frontend/src/app/event/event-filter/event-filter.pipe.ts
```diff
@@ -45,3 +45,18 @@ export class EventFilterPipe implements PipeTransform {
...
+export class EventUserRegistrationStatusPipe implements PipeTransform {
+  transform(statuses: UserRegistrationStatus[], event_id: number): boolean {
+    let filteredStatuses = statuses.filter((status) => {
+      return status.event_id == event_id;
+    });
+
+    return filteredStatuses.length > 0
+      ? filteredStatuses[0].is_registered
+      : false;
+  }
```
KrisJordan, line 61: Seeing this complexity carried forward to the front-end, I am increasingly feeling like doing more heavy lifting and specificity in the backend (e.g. having an endpoint that returns event models that are subclasses of event with user-specific registration field added) will make the front-end implementation require less gymnastics.

KrisJordan (changes requested):
This is looking great! I'm requesting a few changes and I believe the biggest quality of life improvement is going to be around trying to rethink API endpoints just slightly as covered in notes. If this doesn't make sense, let's jump on a call!

ajaygandecha:
@jadekeegan, changing the name of this ticket to be more all-encompassing of the backend *and* frontend work!

jadekeegan:
@KrisJordan Added the registration limit frontend support, registered users list, and unit tests for all new service methods added!

To view the registered users list, go to the event details page at the url `/events/:event_id`. I am hoping to move this to its own "Event Admin" page (along with the editor and other event management features!) once Ajay's gear icon PR comes through.

I have also begun setting up the Profile Page to display a user's registered events, but I am also waiting for Audrey's event pagination PR to be approved so I can use that. :)

backend/main.py
```diff
@@ -85,12 +93,15 @@ def resource_not_found_exception_handler(
...
+feature_exception_handlers = [coworking.exception_handlers, events.exception_handlers]
 
 for feature_exception_handler in feature_exception_handlers:
     for exception, handler in feature_exception_handler:
 
         @app.exception_handler(exception)
-        def _handler_wrapper(request: Request, e: exception):
+        def _handler_wrapper(
+            request: Request, e: ReservationException | EventRegistrationException
+        ):
```
KrisJordan: I'm going to push a minor fix here that uses `Exception` as the common supertype rather than unioning the subtypes. Not sure why it was ever lowercase `exception`, but that's almost certainly on me 🤡

backend/services/event.py
```diff
@@ -189,3 +242,459 @@ def delete(self, subject: User, id: int) -> None:
...
+
+        event_registration_entities = (
+            self._session.query(EventRegistrationEntity)
+            .where(EventRegistrationEntity.event_id == event.id)
+            .all()
+        )
+
+        return [entity.to_model() for entity in event_registration_entities]
+
+    def set_event_organizer(
```
KrisJordan, line 337: Is there a way of controlling this from the front-end? Something we should add a ticket for, if not! Would be great to be able to elevate someone from the registrations list (e.g. for making a workshop lead an organizer).

frontend/src/app/event/event.resolver.ts
```diff
@@ -32,8 +32,12 @@ export const eventDetailResolver: ResolveFn<Event | undefined> = (
       location: '',
       description: '',
       public: true,
+      registration_limit: 250,
```
KrisJordan, line 35: Is this default ever used? Should we default to 0 instead?

frontend/src/app/event/widgets/event-users-list/event-users-list.widget.html
```diff
@@ -0,0 +1,44 @@
...
+      <mat-card-title>Registrations</mat-card-title>
+    </div>
+
+    <!-- Buttons -->
+    <!-- Placeholder template for future registration actions (mass email reminders, delete all, etc.) -->
+    <div class="registrations-actions">
+      <button mat-icon-button>
+        <mat-icon>mail</mat-icon>
+      </button>
+    </div>
```
KrisJordan: We should remove this for production purposes; clicking on the email icon was so tempting and ultimately sad. 😭 

I'll do this in my cleanup commit.

frontend/src/app/event/event-editor/event-editor.component.ts
KrisJordan, line 1: We need to be able to control whether an event is able to be registered for by end users, as well as a registration limit, from the event editor.

KrisJordan: Clicking delete from the editor page pops you back to the main listing page before you get a chance to confirm the deletion.

KrisJordan (changes requested):
Just one major requirement that seems to be missing (below) with a few additional suggestions (in comments).

We need to control `registration_limit` and `can_register` from the `EventDetails` editor.

Of the comments that aren't needed for deploy, but would be nice to have, we also could really use the ability to add a user as the event organizer from the event editor.

backend/main.py
KrisJordan: Pushed this fix.

frontend/src/app/event/widgets/event-users-list/event-users-list.widget.html
KrisJordan: Pushed this fix.

backend/services/event.py
```diff
@@ -61,34 +105,41 @@ def create(self, subject: User, event: Event) -> EventDetails:
         # Ensure that the user has appropriate permissions to create users
         self._permission.enforce(
             subject,
-            "organization.events.manage",
+            "organization.events.*",
```
KrisJordan, line 108: I think by convention, this should either mirror or closely mirror the service method name.

In this case, I think `"organization.events.create"` would be appropriate. Similarly for others. In actual service permission enforcement, the action should probably/never have a `*`, though there may be cases where the object could (failing to think of one right now, though I think I used this before).

KrisJordan (changes requested):
Noticing one pretty significant information leak we need to find a resolution for (and maybe a test to be sure we do not accidentally do so in the future):

Right now, from Docs/API, without being authenticated, I can use the `/api/events` as well as the `/api/events/range` end-points and see the list of registered attendees' profile information. Registrant information should be limited to organizers/admins.

jadekeegan:
Discussed with Ajay how you handled a similar feature to registrations/organizers in the Academics feature and ended up heavily refactoring the registration backend to follow that convention. This resolves the registered users information leak, and it also simplifies the backend service methods as requested!

KrisJordan:
Apologies for being obtuse here, but sounds like you're ready to move toward deployment?

jadekeegan:
> Apologies for being obtuse here, but sounds like you're ready to move toward deployment?

In terms of registration requirements, yes as long as there doesn't seem to be any issues. However, I'm actually almost done with the ticket to add organizer management options and will finish it up today, so I'd rather get that in this PR since it's technically related !

jadekeegan:
@KrisJordan Adding/removing organizers is now live! Ready for another CR (and hopefully ready for deployment if there are no other concerns) whenever you get the chance.

KrisJordan (approved):
LGTM let's merge with a few follow-ups I'll add to an issue! Great work 👏

PR #217, opened 2023-12-22
Implement Course Catalog and Section Offerings Frontend

ajaygandecha:
This major PR provides the implementation of the Angular frontend for the base Course Catalog and Offerings features. This includes all components, services, resolvers, and helper TypeScript data models. This feature will ultimately enable students on the CSXL site to explore courses offered from the CS Department, as well as sections offered in various terms.

## Major Backend Changes
* Added a new table to model the relationship between course sections and rooms.
* Changed the `coworking__room` table to `room` table.
* Expose room data for each course section.
* Keep track of credit hours data for courses.

### Academics Tab
* Add the academics tab at `/academics` on the sidebar to include the hub for all academics features of the CSXL site. This tab is currently blank but incorporates the new Academics module.

### Courses Catalog Page
* Add the course catalog page at `/academics/catalog` to showcase all COMP courses at UNC.
* This data is displayed in a table. Course descriptions are displayed on a dropdown for each table row.

### Section Offerings Page
* Add the section offerings page at `/academics/offerings` to showcase all COMP sections for each term.
* This data is displayed in a table. Course descriptions are displayed on a dropdown for each table row.
* Users can select a term from the dropdown in the top right to discover all sections in a specific term.

## Testing
These changes were thoroughly tested. First, I recreated and repopulated the database. Then, all features were tested on the frontend by running `honcho start`.

All backend tests pass with 100% coverage even after the backend changes were applied.

## Future Considerations
* It might be a good idea to model credit hours per course and add it to the database for each course. 
* Currently, the `/academics` page is blank with just two buttons linking to the course catalog and offerings pages. This will need to be expanded.
* The next branch to be merged into `feature/courses` after this one will be one for admin pages for `Terms`, `Courses`, `Sections`, and `Rooms`.
* We may want a separate `Academics` page specifically for unauthenticated users.


backend/entities/courses/section_entity.py
```diff
@@ -4,6 +4,8 @@
 from sqlalchemy import Integer, String, DateTime, ForeignKey
 from sqlalchemy.orm import Mapped, mapped_column, relationship
 
+from backend.models.room_assignment_type import RoomAssignmentType
```
KrisJordan, line 7: Should we prefer relative import paths? (Nit)

frontend/src/app/academics/courses.models.ts
KrisJordan, line 1: Thoughts on whether we should move to a class / interface per file standard? I see trade offs both ways.

frontend/src/app/academics/courses.models.ts
```diff
@@ -0,0 +1,55 @@
+import { TimeRange } from '../coworking/coworking.models';
```
KrisJordan, line 1: It would be great to lift TimeRange to top-level, like you did with Room.

KrisJordan (changes requested):
This is awesome! A few minor nits, one larger request for elevating TimeRange, and one more open question: academics module in frontend vs courses in backend. I think we should aim for parity here, want to see if vscode can handle the refactoring of courses to academics in the backend?

frontend/src/app/academics/courses.models.ts
ajaygandecha: @KrisJordan agree! I see benefits to both. I think it would be easier to find files, however I do see issues with an even more complicated file structure than we already have! I suppose a tradeoff would be putting everything in a `/models` subdirectory like we have with `/widgets`.

Thoughts?

backend/entities/courses/section_entity.py
ajaygandecha: Agree @KrisJordan! We can do this - automatically importing using VSCode uses absolute paths, but I try to convert them where I can.

ajaygandecha:
@KrisJordan great point about `academics` on the backend! I was thinking about this as well a little bit. I was wondering to what extent academics data would be extended. For example, if we add course scheduling, office hours, etc, would all of it be incorporated into academics?

I think, ultimately though, it probably is best practice here because having the module named `courses` when `course` is an item of `courses` does seem a bit wrong! Will make the refactor 😄

ajaygandecha:
Will also rename all tables from `courses__*` to `academics__*`.

ajaygandecha:
Will also refactor all APIs to use the `/api/academics` endpoint instead of `/api/courses` for the same reason.

ajaygandecha:
@KrisJordan done! Reopening for review.

frontend/src/app/academics/courses.models.ts
KrisJordan: Let's leave as is for now.

KrisJordan (approved):
Looks great, thanks for going through the exercise of refactoring around academics! Lgtm!

PR #219, opened 2023-12-24
Event Pagination

atoneyd:
This PR implements pagination for the events tab. The page defaults to show upcoming events for the next month, with buttons for selecting previous or next months. You can also search all events, as long as your search is > 2 characters and searches are also limited through debouncing.

- Added the `EventPaginationParams` model, which includes pagination parameters for specifying a start and end date, ascending/descending, order by, and filter
- Added two more events as demo data for testing this feature
### API Changes
Exposes the following API to the Angular Frontend via FastAPI:
- GET `api/events/paginate`: returns a paginated list of events based on the specified pagination parameters
### Service Changes
- Added the `list()` function to the backend `EventService` to query and paginate events based on the parameters passed in from the API call
### Frontend
- Added corresponding frontend model and service function
- Updated EventPageComponent to include the buttons for navigating between the previous and next months of events
- Updated search bar on this page to include debouncing and send search requests through API calls rather than filtering on the frontend
- Remove eventFilterPipe now that filtering is done on the backend

Picture:
image

## Testing
Backend: Added a new PyTest to test the list() function. Confirm 100% passing by running `pytest`
Frontend:
- Run `honcho start` and `python3 -m backend.script.reset_demo` 
- Navigate to the Events tab and see that the default dates are set to `Today - One Month from Today`. There should be four events listed in ascending order
- Select Previous Month, and date range at the top should change to last month. There should be two events listed in descending order
- Search for 'ca.' Nothing should happen. Now search for 'cads'. Four events should appear. Clearing the search bar should reset data to the correct date paginated view.

## Future Considerations
- Put pagination stuff into a widget for reusability, if that is possible. Wasn't sure how to to neatly and properly do that here since there were interactions with routes and other things.
- Note: I am no longer using the eventResolver; I was not sure if I should delete it or change my code to incorporate it back in


frontend/src/app/event/event-page/event-page.component.html
```diff
@@ -3,14 +3,24 @@
...
     <div class="events-grid">
       <div class="events-column">
         <!-- `eventsPerDay` is in the format of a list of tuples, where
                     eventGroup[0] is the date string and eventGroup[1] s a list of events -->
+        <div class="date-range">
+          <button mat-stroked-button (click)="showEvents(true)">
+            Previous Month
+          </button>
+          {{ startDate | date: 'shortDate' }} -
+          {{ endDate | date: 'shortDate' }}
```
KrisJordan, line 19: I would love to see a better representation up here, but this is non-blocking and can be fixed after deploy. Some custom logic may be needed for like "Today - Jan 26th, 2024" vs. "Jan 26th - Feb 26th, 2024"

Another issue to file for later, we should try to use the start minute/hour/second as first of the morning on start date and last of evening on end date, so that next pages always go to the next day.

frontend/src/app/event/event-page/event-page.component.ts
```diff
@@ -91,7 +180,26 @@ export class EventPageComponent implements OnInit {
...
    */
   onSearchBarQueryChange(query: string) {
-    this.eventsPerDay = this.eventService.groupEventsByDate(this.events, query);
+    this.query = query;
+    let paginationParams = this.page.params;
+    paginationParams.ascending = 'true';
+    if (query == '') {
+      console.log('hi');
+      paginationParams.range_start = this.startDate.toLocaleString('en-GB');
+      paginationParams.range_end = this.endDate.toLocaleString('en-GB');
```
KrisJordan, line 189: This should really be en-US, yeah? Or is there something I'm missing about great britain representation.

frontend/src/app/event/event-page/event-page.component.ts
```diff
@@ -91,7 +180,26 @@ export class EventPageComponent implements OnInit {
    * @param query: Search bar query to filter the items
    */
   onSearchBarQueryChange(query: string) {
-    this.eventsPerDay = this.eventService.groupEventsByDate(this.events, query);
+    this.query = query;
+    let paginationParams = this.page.params;
+    paginationParams.ascending = 'true';
+    if (query == '') {
+      console.log('hi');
```
KrisJordan, line 187: I'm going to remove this.

frontend/src/app/event/event-page/event-page.component.ts
KrisJordan: I'm going to try pushing this fix, it doesn't seem to be breaking anything on my end, but could use a sanity check on your end if there are any cases I'm not thinking of.

KrisJordan (approved):
This looks really great, I'm going to go ahead and approve! Really impressed with where this UX landed, it feels great to see it all come together. @atoneyd please confirm that changing to `en-US` should not break anything. Once confirmed, merge at will.

frontend/src/app/event/event-page/event-page.component.ts
KrisJordan: Removed.

ajaygandecha:
@atoneyd / @jadekeegan / @KrisJordan, with #276 open, I will close this PR. Great work getting this started !!

PR #220, opened 2023-12-24
Implement Admin Features for the Academics Feature

ajaygandecha:
This major PR provides the implementation of the Angular frontend and Python backend for the academic admin features. This includes all components, services, resolvers, and helper TypeScript data models. This feature will ultimately enable CSXL administrators to add, update, and delete courses, sections, terms, and rooms.

## Major Backend Changes
* Edit some backend services in `/academics` to better support admin features.
* Add a `RoomService` to implement CRUD operations for the `room` table.
* Add necessary APIs to expose `RoomService` functionality.

### Navigation Component Refactor for Admin Page Link

This PR adds a gear icon to the top right of the Navigation toolbar exposing the admin page to users with the correct permissions. This gear icon links to the admin page.

To implement this, a new frontend service called the `NagivationAdminGearService` manages when to show the gear. Upon redirect, the navigation component clears gear data, and on initialization, components use the `NagivationAdminGearService.showAdminGear(permissionAction: string, permissionResource: string,  tooltip: string, targetUrl: string)` to conditionally show the gear on the navigation bar if the permissions are met.

This feature can easily be added throughout the CSXL application. For now, the functionality is only used in the academics admin features.

### New Components

This PR adds nine new components to handle academics admin features, including:
* **Academics Admin Home**: Exposes the academics admin features.
* **Term Admin**: Shows all term data and exposes CRUD functionality.
* **Course Admin**: Shows all course data and exposes CRUD functionality.
* **Section Admin**: Shows all section data and exposes CRUD functionality.
* **Room Admin**: Shows all room data and exposes CRUD functionality.
* **Term Editor**: Form to show when terms are to be created and edited.
* **Course Editor**: Form to show when courses are to be created and edited.
* **Section Editor**: Form to show when sections are to be created and edited.
* **Room Editor**: Form to show when room are to be created and edited.
 
The main admin pages for each feature also have their own respective `RxObject` implementation for live refreshing of data upon deletion.

## Testing
These changes were thoroughly tested. First, I recreated and repopulated the database. Then, all features were tested on the frontend by running `honcho start`.

All backend tests pass with 100% coverage even after the backend changes were applied.

## Future Considerations
* If we begin to add more course types to the page, I would love to switch the input select for course subject codes to use the material chip components.
* We can now implement the gear icon for other admin features and refactor the folder structure - notably, for organizations.
* We may want a separate `Academics` page specifically for unauthenticated users.
* We can consider creating detail pages for courses and terms. At the moment though, it does not seem necessary.


backend/api/academics/section.py
```diff
@@ -98,7 +98,15 @@ def new_section(
     Returns:
         SectionDetails: Section created
     """
-    return section_service.create(subject, section)
+    created_section = section_service.create(subject, section)
+    section_service.add_lecture_room_to_section(
+        subject, created_section, section.lecture_room
+    )
+
```
KrisJordan, line 105: Without looking at the implementation of these service methods, I'm a little surprised and slightly confused by how this API implementation reads.

I wonder if a little refactoring or commenting could make it more obvious why the mutating call is needed after create and why the conditional statement is needed?

backend/api/room.py
```diff
@@ -0,0 +1,123 @@
...
+api = APIRouter(prefix="/api/room")
+openapi_tags = {
+    "name": "Rooms",
+    "description": "Create, update, delete, and retrieve rooms.",
+}
+
+
+@api.get("", response_model=list[RoomDetails], tags=["Rooms"])
+def get_rooms(
+    room_service: RoomService = Depends(),
```
KrisJordan, line 26: Room Service is a fun service name 😄

backend/api/room.py
```diff
@@ -0,0 +1,123 @@
...
+    Get all room
+
+    Parameters:
+        room_service: a valid RoomService
+
+    Returns:
+        list[RoomDetails]: All rooms in the `Room` database table
+    """
+
+    # Return all organizations
```
KrisJordan, line 38: Left over comment from orgs

KrisJordan (approved):
Looks great, there is just one place I found slightly confusing and another small nit. Exciting to see this coming together! Going to be great models which many other features of the site will take dependencies on.

backend/api/academics/section.py
ajaygandecha: @KrisJordan yeah, this was a very interesting issue that I was discovering as well. This stems from the fact that our `SectionModel` splits up `rooms` into two properties -- `lecture_room` and `office_hour_rooms`. So, when we try and POST a section model, we cannot post directly into the table for the section entity.

My solution was this super ugly implementation.. ☹️ - it relies on section being created *first* (so that its ID field is populated), then connecting it to a room in the `academics__section_room` table.

Do you think this is alright here? I will add a comments if so.

ajaygandecha:
@KrisJordan, added the documentation!

ajaygandecha:
Going to go ahead and merge this in since changes were approved, that way I can send this up for final review! 🥳 More documentation and a slight rewrite was done on the API code. Feel free to bring up again in the overall review if there is a better way! I tried to combine everything in the service only twice, and for some reason, was not having any luck.

PR #224, opened 2023-12-29
Minor changes for the academics feature

KrisJordan:
1. Added the view only `staff` relationship to `SectionEntity` which places a limit on the join.
2. Reverted my suggested enum changes because after looking in the actual database, SqlAlchemy appears to only store the enum name not the enum value.
3. Fixed some imports in coworkign models that prevented the repl script from loading
4. Slight modification to the current term query to limit the number of rows returned by the database to be 1 in SQL rather than in fetching after SQL executes

### DIFF
--- a/backend/entities/academics/section_entity.py
+++ b/backend/entities/academics/section_entity.py
@@ -54,6 +54,13 @@ class SectionEntity(EntityBase):
         back_populates="section"
     )
 
+    # Relationship subset of members queries for non-students
+    staff: Mapped[list["SectionMemberEntity"]] = relationship(
+        back_populates="section",
+        viewonly=True,
+        primaryjoin="and_(SectionEntity.id==SectionMemberEntity.section_id, SectionMemberEntity.member_role!='STUDENT')",
+    )
+
     @classmethod
     def from_model(cls, model: Section) -> Self:
         """
@@ -89,11 +96,7 @@ class SectionEntity(EntityBase):
             for room in self.rooms
             if room.assignment_type == RoomAssignmentType.OFFICE_HOURS
         ]
-        staff = [
-            members.to_flat_model()
-            for members in self.members
-            if members.member_role != RosterRole.STUDENT
-        ]
+        staff = [members.to_flat_model() for members in self.staff]
 
         return Section(
             id=self.id,
--- a/backend/entities/academics/section_entity.py
+++ b/backend/entities/academics/section_entity.py
@@ -96,7 +96,6 @@ class SectionEntity(EntityBase):
             for room in self.rooms
             if room.assignment_type == RoomAssignmentType.OFFICE_HOURS
         ]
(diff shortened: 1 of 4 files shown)


backend/entities/academics/section_entity.py
```diff
@@ -54,6 +54,13 @@ class SectionEntity(EntityBase):
         back_populates="section"
     )
 
+    # Relationship subset of members queries for non-students
+    staff: Mapped[list["SectionMemberEntity"]] = relationship(
+        back_populates="section",
+        viewonly=True,
+        primaryjoin="and_(SectionEntity.id==SectionMemberEntity.section_id, SectionMemberEntity.member_role!='STUDENT')",
```
ajaygandecha, line 61: Ah, so this was the solution!!! Very interesting - thanks for including this @KrisJordan !

backend/entities/academics/section_entity.py
```diff
@@ -89,11 +96,7 @@ def to_model(self) -> Section:
             for room in self.rooms
             if room.assignment_type == RoomAssignmentType.OFFICE_HOURS
         ]
-        staff = [
-            members.to_flat_model()
-            for members in self.members
-            if members.member_role != RosterRole.STUDENT
-        ]
+        staff = [members.to_flat_model() for members in self.staff]
```
ajaygandecha, line 99: Maybe can move back into the Section() constructor for conciseness?

ajaygandecha (approved):
LGTM @KrisJordan! I went ahead and fixed the one small nit I included in a commit. 

Before we merge, do we want to make the entity field change for the lecture room and office hour rooms as well?

ajaygandecha:
@KrisJordan In addition, with this, I wonder if adding a field for students here (without editing models so this data isn't exposed yet) might be useful so that we don't need a migration for this later

backend/entities/academics/section_entity.py
KrisJordan: Might as well use in those instances, too, since the solution is the same and it is straightforward. It's too bad this gets jammed into a string and evaled rather than being straight source from the start, but I was going off SQLAlchemy example

backend/entities/academics/section_entity.py
ajaygandecha: (Resolved in latest commit btw)

KrisJordan:
> @KrisJordan In addition, with this, I wonder if adding a field for students here (without editing models so this data isn't exposed yet) might be useful so that we don't need a migration for this later 

I could see this making sense, but I think it's generally best to avoid the temptation of building out code that will lie dormant in anticipation of being hooked up later. The good news is adding a view only relation is in Python only and will not need a migration generated.

backend/entities/academics/section_entity.py
ajaygandecha: ... I realized I never pushed. Did now

PR #228, opened 2023-12-31
Temporarily Hide Instructor Column on Section View

ajaygandecha:
This small PR temporarily hides the 'Instructor' column on the Section Offerings view while the admin feature for staff is developed.


KrisJordan (approved):
lgtm

PR #229, opened 2023-12-31
Improved DX scripts:  delete/create database in reset scripts, add migration generation script

KrisJordan:
This PR has two developer experience (DX) quality of life improvements:

1. Improves upon the reset_demo and reset_test backend scripts by shelling out to run the scripts database_delete and database_create each time a database is reset. This avoids a common issue when developing new schema with foreign key constraints where resetting fails, so developers would need to manually go back and run delete/create database scripts.
2. Adds a script for generating migrations based on a remote branch.

### DIFF
--- a/backend/script/generate_migration.py
+++ b/backend/script/generate_migration.py
@@ -59,11 +59,11 @@ def main() -> None:
         print("❌ Failed to switch to main branch.")
         sys.exit(1)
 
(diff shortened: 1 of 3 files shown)


jadekeegan (approved):
LGTM! New scripts work as expected. The migration script is very cool!

PR #234, opened 2023-12-31
Event Organizer Fixes #233

### LINKED ISSUE #233: Event Organizers Fixes
There are two issues outstanding on Event Organizers:

Screenshot 2023-12-31 at 1 31 04 PM

- [x] The chip profile picture shows the current user's profile picture rather than the selected users (example above)
- [x] On the Registrations widget, Event Organizers are shown without any indication that that's who they are. Perhaps we add some kind of indicator (a ⭐ ?) next to their name and bold them with a *. It may be worth collapsing first/last name into a single Name column? Alternatively, don't show organizers in Registrations, but do show them in the event details card beneath the description (with contact information).

jadekeegan:
This PR improves the UI/UX as it relates to Event Organizers by fixing the Github avatar bug, adding Event Organizers to the Event Details card, removing them from the registrations list, and fixing the Event Organizer form control on the Event Editor form so it properly implements the remove organizers button.

> This PR resolves issue #233.

### DIFF
--- a/backend/entities/event_registration_entity.py
+++ b/backend/entities/event_registration_entity.py
@@ -115,9 +115,10 @@ class EventRegistrationEntity(EntityBase):
         """
         return EventOrganizer(
             id=self.user_id,
+            registration_type=self.registration_type,
             first_name=self.user.first_name,
             last_name=self.user.last_name,
             pronouns=self.user.pronouns,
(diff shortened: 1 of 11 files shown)


KrisJordan (approved):
LGTM!

Issue #232, opened 2023-12-31
Remove `can_register` boolean column from `EventEntity`


KrisJordan:
Does `EventEntity#can_register` need to exist at the entity level? It seems to be a computed field that is a projection of `registration_limit > 0`?

PR #235, opened 2023-12-31
Remove can_register Column from Event Entity #232

jadekeegan:
This PR removes the `can_register` boolean column from the `EventEntity` since it is a computed field based on the `registration_limit`.

> This PR resolves issue #232.


KrisJordan (approved):
These changes look great! Awesome to see the migration in there, as well! Reviewed source changes on phone, will give a quick validation, merge, and deploy/migrate when back at computer shortly!

PR #236, opened 2024-01-01
Improve Event Detail Card Styling

jadekeegan:
This PR fixes the styling issues with the organizer chip avatars on the Event Details card. It also adds a scroll bar to the event details card so information is still viewable if the text overflows the card.


KrisJordan (approved):
LGTM

PR #237, opened 2024-01-02
Convert Reusable Code to Widgets & Refactor Event/Organizer Models

jadekeegan:
This PR converts code for a user look up form field as well for a list of users as chips which was previously in individual components into widgets. It also creates a `PublicUser` model to return user information without exposing sensitive user details.

## Major Changes
- Created the `user-chip-list` widget.
- Created the `user-lookup` widget.
- Removed the `EventOrganizer` and `EventMember` models and replaced with `PublicUser` (backend)/`PublicProfile` (frontend) model.
- Removed the `attendees` field in the Event model as it was not used and to ensure that no attendee information is exposed.
- Resolved chip alignment and light mode chip text CSS bugs.


KrisJordan (approved):
LGTM

PR #244, opened 2024-01-26
Ng serve polling fix

### LINKED ISSUE #78: Make "--poll 2000" System Specific
# Make "--poll 2000" System Specific
Create a secondary script through `npm` that runs the correct `ng serve` command based on OS.

## Why?
This is necessary to ensure that the local server automatically refreshes on Windows. However, it is suboptimal and not necessary for Mac OS systems.

## Acceptance Criteria
- [ ] Create some script to determine what system the user is using and add `--poll 2000` accordingly.

noahsmiths:
These changes implement a mechanism to differentiate between windows and non-windows systems from within the devcontainer. This is done through the use of the default `windir` (case sensitive) environment variable on windows, which is then forwarded to the devcontainer through the docker-compose file if it exists.

A new npm script is included called `start:dynamic` which will append the `--poll 2000` flag to `ng serve` iff the `windir` variable exists in the container, which will only occur on windows machines.

Closes #78

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -10,7 +10,7 @@ services:
       - csxl.unc.edu-node_modules:/workspace/frontend/node_modules
     command: /bin/sh -c "while sleep 1000; do :; done"
     environment:
-      - OS # Defined on Windows but not on other platforms
+      - WINDIR # Defined on Windows but not on other platforms
   db:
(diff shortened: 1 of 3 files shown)


jadekeegan (approved):
LGTM! Great work solving this long overdue issue!!

@KrisJordan Ready to merge.

PR #245, opened 2024-01-26
230 click  pointer affordance of academics admin tables

### LINKED ISSUE #230: Click / Pointer Affordance of Academics Admin Tables
When viewing the admin tables of Academics Administration, the hand pointer appears in the black space of a row of a section/course/room/term, however, clicking in this blank space (which has the hand pointer CSS mouse icon) results in no action being taken.

Desired fix: show some kind of "hover" state (border change, background change) for a given row when in a clickable area and the hand pointer only in places where you can click and an action is taken. Ideally, you should be able to click anywhere left of the actions column with the delete button and it takes you to the editor. Currently you must click on the name itself.

Not super high priority, but a good first issue.

wherthey:
closes #230 

This branch allows an admin to click anywhere in the desired row when trying to edit a section, course, room, and term. The issue mentioned adding a hover state, but that already existed. If you would like that state to be something different, please let me know and i will change it.

### DIFF
--- a/frontend/src/app/academics/academics-admin/course/admin-course.component.html
+++ b/frontend/src/app/academics/academics-admin/course/admin-course.component.html
@@ -9,12 +9,12 @@
         </div>
       </th>
       <td mat-cell *matCellDef="let element">
-        <div class="row">
-          <p (click)="updateCourse(element)">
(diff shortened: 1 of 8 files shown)


KrisJordan (approved):
LGTM, great work

PR #248, opened 2024-01-28
Community Agreement

tmattvu:
- Added has_agreed field for user and profile models on frontend and backend to keep track of whether or not they accepted the community agreement

- Updated necessary API call for user

- Added community agreement widget to shared folder

- FOR EXISTING USERS: When accessing home page of the csxl website, a pop up will show up with the agreement asking to accept if the user has not accepted yet

- Added text on top of pop up box notifying user if they accepted agreement yet

- Copy of community agreement can be found on 'About the XL' page

- Added pop up agreement for new user registration

- Pop up only has one button (accept or close)


KrisJordan:
This is looking really great; would love a demo at today's meeting!

Since this has a schema change, we will need to be careful with deployment and migration. I can talk through migration process today.

One thing I'd like to discuss today is whether `has_agreed` should be a version number (e.g. community agreement 1.0) as things may change in the future. I suppose we could also write a script that resets everyone back to False here and get a similar outcome (but erases that they agreed at some point in the past).

I'd also like to discuss whether `has_agreed` is the right name here. While much more verbose, `accepted_community_agreement`, or something else, maybe even just `community_agreement` is a little more obvious to someone coming in and reading the fields without context.

tmattvu:
Sounds good. After the meeting today, I'll go back and change the has_agreed field to accepted_community_agreement!

tmattvu:
> This is looking really great; would love a demo at today's meeting!
> 
> Since this has a schema change, we will need to be careful with deployment and migration. I can talk through migration process today.
> 
> One thing I'd like to discuss today is whether `has_agreed` should be a version number (e.g. community agreement 1.0) as things may change in the future. I suppose we could also write a script that resets everyone back to False here and get a similar outcome (but erases that they agreed at some point in the past).
> 
> I'd also like to discuss whether `has_agreed` is the right name here. While much more verbose, `accepted_community_agreement`, or something else, maybe even just `community_agreement` is a little more obvious to someone coming in and reading the fields without context.

Fixed field naming!

frontend/src/app/shared/community-agreement/community-agreement.widget.html
```diff
@@ -0,0 +1,196 @@
+<mat-card class="community-agreement-dialog" #dialogContainer>
+  <mat-card-content class="content-container">
+    <p *ngIf="has_user_agreed">
```
jadekeegan, line 3: NIT: Maybe remove this text for `has_user_agreed`  and update  the `!has_user_agreed`  <p> with text that says something like "Please read and accept the community agreement below to continue to the CSXL web page." Make sure to align it with the rest of the text in the card though!

frontend/src/app/shared/community-agreement/community-agreement.widget.html
```diff
@@ -0,0 +1,196 @@
...
+        Science Undergraduate Community at UNC-Chapel Hill.
+      </p>
+      <p>
+        The XL is dedicated to providing a harassment-free and inclusive
+        experience for everyone regardless of gender identity and expression,
+        sexual orientation, disabilities, neurodiversity, physical appearance,
+        body size, ethnicity, nationality, race, religion, or other protected
+        category. These policies are inspired by and almost entirely replicate
+        Google's Developer Community Guidelines:
+        https://developers.google.com/community-guidelines
```
jadekeegan, line 27: IMPT: Update this and other links so they are clickable! Maybe also change it so that the link text doesn't appear and you just click the text like "Google's Developer Community Guidelines" to go to the link since the links run off the card on mobile.

frontend/src/app/shared/community-agreement/community-agreement.widget.html
```diff
@@ -0,0 +1,196 @@
+<mat-card class="community-agreement-dialog" #dialogContainer>
```
jadekeegan, line 1: NIT: Could take this html out of the card so the dialog doesn't appear as a "card within a card".

frontend/src/app/shared/community-agreement/community-agreement.widget.css
```diff
@@ -0,0 +1,11 @@
+.community-agreement-dialog {
```
jadekeegan, line 1: NIT: Not necessary, but if you want to, you could bring over the scrollbar styling from the Event Details Card! The default scrollbar is just a little clunky compared to the style of the XL site.

jadekeegan (changes requested):
This is looking really great so far!! I added some comments in the code for a handful of changes. Most are nitpicks, but I think making links in the agreement clickable is fairly important.

The only other major change request I have is that if you click outside of the community agreement form, it automatically closes it and allows the user full access to coworking without them accepting it.

frontend/src/app/shared/community-agreement/community-agreement.widget.html
tmattvu: Fixed!

frontend/src/app/shared/community-agreement/community-agreement.widget.html
tmattvu: Fixed

frontend/src/app/shared/community-agreement/community-agreement.widget.html
tmattvu: Fixed!

frontend/src/app/shared/community-agreement/community-agreement.widget.css
tmattvu: Done!

tmattvu:
Fixed all requested changes! 

FYI: User can no longer click outside the pop up box to close it when they first register or when it pops up on coworking home page. However, users can click outside the pop up box to close it on the about the xl page (which would only be able to be seen after a user has already accepted the agreement)

tmattvu:
> One additional place where I would suggest perhaps checking this, for completeness, is in the backend route for drafting a reservation in the CSXL. After checking for permissions, perhaps check that if the agreement is not accepted we raise an exception that causes the request to draft a reservation to fail.

Done! Added the following line to draft_reservation function in backend! #
```python
if subject.accepted_community_agreement == False:
            raise ReservationException("User has not accepted community agreement.")
``` 

Also added testing for it and updated all the other tests in the draft_res file to account for new error thats thrown since user data has accepted_community_agreement set to False.

jadekeegan (approved):
LGTM!! Great work! Next step is migration. I believe Kris showed the document, but if you have any issues/questions, let me know. I'll review the migration script again once it's made.

frontend/src/app/coworking/coworking-home/coworking-home.component.ts
KrisJordan, line 1: This is feeling out of place; even though it's what we described. What if we moved this to `navigation.component.ts`; there we already have a handle on `ProfileService` via `$profile`. So we would need to add another subscriber to check and see if the user had accepted and force it once logged in with a profile. The only potential concern with this route is being sure new user registration still works.

frontend/src/app/profile/profile-editor/profile-editor.component.ts
```diff
@@ -78,6 +81,14 @@ export class ProfileEditorComponent implements OnInit {
   onSubmit(): void {
     if (this.profileForm.valid) {
       Object.assign(this.profile, this.profileForm.value);
+      if (this.profile.accepted_community_agreement === undefined) {
```
KrisJordan: Going to change this to `!this.profile.accepted_community_agreement` which handles the case of undefined but also handles the case of `false`, which is actually more expected based on the type of `profile` being `Profile`.

frontend/src/app/shared/community-agreement/community-agreement.widget.ts
```diff
@@ -0,0 +1,60 @@
...
+  onCloseClick(): void {
+    this.dialogRef.close();
+  }
+
+  onAcceptClick() {
+    this.profileService.profile$.subscribe((profile) => {
+      if (profile) {
+        profile.accepted_community_agreement = true;
+        const url = '/api/profile';
+        this.http.put(url, profile).subscribe();
```
KrisJordan, line 55: This logic should go in `profileService` (and already should be there with profile updating...); this widget should not be concerned with `HttpClient` to keep the layers of abstraction consistent.

frontend/src/app/shared/community-agreement/community-agreement.widget.ts
```diff
@@ -0,0 +1,60 @@
+/**
+ * The Community Agreement Widget showcases the agreement that existing and new users
+ * will have to sign. It is currently used on the coworking home page and the 'About the
+ * XL' page. It also checks whether or not users have accepted.
+ *
+ * @author Matt Vu
+ * @copyright 2023
```
KrisJordan, line 7: 2024!

frontend/src/app/shared/community-agreement/community-agreement.widget.html
```diff
@@ -0,0 +1,209 @@
+<mat-card-content class="content-container">
+  <p *ngIf="!has_user_agreed" style="margin-left: 15px; margin-top: 10px">
+    ***Please read and accept the community agreement below to continue to CSXL
+    web page.***
```
KrisJordan, line 4: This currently looks a little half-baked with the asterisks in paragraph tags. Can we elevate this to be the header of the dialog using the `mat-dialog-header`: https://material.angular.io/components/dialog/examples#dialog-content

frontend/src/app/shared/community-agreement/community-agreement.widget.html
```diff
@@ -0,0 +1,209 @@
...
+      We do not tolerate harassment of XL community members in any form. We take
+      violations of our policy seriously and will respond appropriately.
+    </p>
+    <p>
+      All participants of the XL in-person and on-line communities, XL
+      ambassadors, XL staff, speakers, faculty, and participants otherwise, must
+      agree to and abide by the following policies in order to access XL's
+      events, online communities, and facilities.
+    </p>
+    <h3>I agree to be respectful to everyone in the XL community.</h3>
```
KrisJordan, line 40: It's currently a little too easy to just blow throw the reading of this agreement like a typical EULA and not read anything. Can we make the case where someone _has not agreed_ include check boxes that must be checked for each of these `h3` elements? In the `Accept` button case, we can check to be sure each has been checked and ask the user to agree to all community expectations if they did not check any.

KrisJordan (changes requested):
Looking great! I pushed a small commit with a few fixes.

There are a few places where some changes are requested in the comments. Additionally, where there are files with copyright, be sure to add yourself and update the dates (2024 for new files, 2023-2024 for revised files).

I'll take a stab at generating the migration, right now.

KrisJordan:
Added the migration script in 1c7afdf757fe60dda1163b8be8abf0ecd18fbe65

backend/test/services/coworking/reservation/draft_test.py
```diff
@@ -45,6 +45,8 @@
 def test_draft_reservation_open_seats(
     reservation_svc: ReservationService, time: dict[str, datetime]
 ):
+    user_data.ambassador.accepted_community_agreement = True
+    assert user_data.ambassador.accepted_community_agreement == True
```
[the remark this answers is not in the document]
tmattvu: Fixed dummy data to have accepted_community_agreement = True and removed those lines in the testing file.

frontend/src/app/about/about.component.html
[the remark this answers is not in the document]
tmattvu: Moved to Profile Page! Hidden until user saves their information like everything else on the page.

frontend/src/app/about/about.component.ts
```diff
@@ -9,4 +12,18 @@ export class AboutComponent {
...
   };
+
+  constructor(
+    protected profileService: ProfileService,
+    protected dialog: MatDialog
+  ) {}
+
+  openAgreementDialog(): void {
+    const dialogRef = this.dialog.open(CommunityAgreement, {
+      width: '1000px',
```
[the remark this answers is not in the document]
tmattvu: The width was there initially before applying css to my widget!

backend/test/services/user_test.py
```diff
@@ -234,3 +234,33 @@ def test_update_user_enforces_permission(
...
+    current_user.last_name = "Samuel"
+    updated_user = user_svc.update(root, current_user)
+    assert updated_user is not None
+    assert updated_user.id == current_user.id
+    assert updated_user.first_name == "Sam"
+    assert updated_user.last_name == "Samuel"
+    assert (
+        updated_user.accepted_community_agreement
+        == current_user.accepted_community_agreement
+    )
```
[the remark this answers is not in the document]
tmattvu: Changed it to where there are now two tests. One where user who hasn't accepted community agreement still has to accept after updating (even though I don't think it is possible because of the frontend) and one where a user does not have to reaccept terms after updating profile.

KrisJordan (approved):
LGTM!

PR #250, opened 2024-01-30
Room Reservations

yuvrajjain2003:
This PR contains the work done by Team B8 in Fall 2023 in COMP 423. Our team primarily worked on implementing the room reservations feature, which should be mostly isolated from the rest of the codebase. 

So I would consider this to be a successful PR as long as it does not cause any errors in the rest of the codebase. You should be able to navigate to our primary land page by navigating to Coworking and clicking on the Room Reservation button. This should display a table which you can interact with to make reservations.


jadekeegan (changes requested):
Finally had the chance to skim through the code and explore on the local server. This feature looks awesome! I'm excited to eventually see it on the XL site!

Aside from the concerns you mentioned regarding administration and reservation limits, here are a some other items I noticed:
1. I feel like there should be a required number of users to reserve a room to ensure that one user can't reserve a 16-person room like SN 136 for themselves. Along with this, you would need to be able to add multiple users to a reservation at a time. This would also prevent a group of people from reserving half the rooms on the hallway or stacking reservations.
2. Aside from Room Reservation admin, we also need to add things to the Room admin page such as a reservable status (since other class rooms won't be included in our reservation system), capacity, reservable times (so we can block out unavailable hours for OH), etc.
3. Some of the navigation feels a bit weird. I think it would be nice to have separate pages for upcoming reservations and just a button to make a new reservation, or doing something in that vein to improve how you find your upcoming reservations. If I just want to view my reservations, it seems unnecessary to have the "New Reservation" form there before that.
4. Need to update the "New Reservation" form to fit with Angular Mat UI and be consistent with the rest of the website in terms of margins and spacing. Using cards, maybe rounding the edges of the key, changing the colors a little bit to match the theme, etc.
5. Generally need to update the UI of the Room Reservation page in a few different ways. My main concern is that this page won't work super well one mobile. Some thoughts I had was a dropdown to choose the room, and then you could maybe make the time selection vertical rather than horizontal. Could also look into removing the text for every single 30 minute increment so it's not as crowded too.
6. 'Reserve' might make more sense rather than 'Select' for the button to create the reservation. Would also be nice to have the Cancel button return back to the reservation form so they can continue editing.
7. 'Confirm' for reservation should return to the reservations page rather than the home page.

Feel free to message me/find me in person if you would like to more thoroughly discuss these things!

Also, I didn't thoroughly go through the code itself. Was mostly looking at things on a higher level, but if I think Ajay said he might do a more technical review soon.

@KrisJordan Any other thoughts on your end?

PR #255, opened 2024-02-05
Replace hardcoded references to /workspace with relative imports/paths

CanNotCode2:
This allows the project to work outside of the docker container environment, removing the need to have a /workspace path for the project to be deployed in.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -9,6 +9,8 @@ services:
       - ..:/workspace
       - csxl.unc.edu-node_modules:/workspace/frontend/node_modules
     command: /bin/sh -c "while sleep 1000; do :; done"
+    environment:
+      - windir # Defined on Windows but not on other platforms
   db:
     image: postgres:15.2
     restart: unless-stopped
--- a/Procfile
+++ b/Procfile
@@ -2,9 +2,9 @@
 #
 # Start/spawn all processes with `honcho start` at the command-line.
 # Stop children processes with Control+C to send the interrupt signal.
-# 
+#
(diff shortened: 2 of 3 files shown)


frontend/.eslintrc.json
```diff
@@ -10,7 +10,7 @@
       ],
       "parserOptions": {
         "project": [
-          "/workspace/frontend/tsconfig.json"
+          "./tsconfig.json"
```
KrisJordan, line 13: This change breaks `eslint`; please validate the PR does not break anything before resubmitting. To validate, open a TS file in `frontend` (such as `src/main.ts`) and ensure that you do not see an error such as `Cannot find file: /workspace/tsconfig.json`. In this case, the editor/eslint is looking relative to project root not frontend. Please fix and push again.

KrisJordan (changes requested):
Thanks for making the PR!

One requested change in comments and one requested change for project standards. The project standards request is to please use a real name and profile photo for contribution recognition purposes. Will merge in after these changes are made, thanks!

frontend/.eslintrc.json
CanNotCode2: I have. It works. `./frontend/tsconfig.json` breaks it. ESLint is searching from relative to where `.eslintrc.json` is located not the project root.

KrisJordan: This was tested in the devcontainer and breaks the devcontainer.

CanNotCode2: I'll change it to ./frontend/tsconfig.json for now then and investigate further to try and find a way to get this to work everywhere.

PR #256, opened 2024-02-06
Feedback Form Feature

tmattvu:
Draft PR to keep track of updates!

### DIFF
--- a/backend/entities/feedback_entity.py
+++ b/backend/entities/feedback_entity.py
@@ -1,0 +1,100 @@
+"""Definition of SQLAlchemy table-backed object mapping entity for Feedback."""
+
+from sqlalchemy import Column, Integer, String, Boolean, ForeignKey
+from sqlalchemy.orm import Mapped, mapped_column, relationship
+
+from backend.models.feedback import Feedback
+from backend.models.feedback_details import FeedbackDetails
+from .entity_base import EntityBase
+from typing import Self
+from ..models.user import User
+
+
+__authors__ = ["Matt Vu"]
+__copyright__ = "Copyright 2023"
+__license__ = "MIT"
+
+
+class FeedbackEntity(EntityBase):
+    """Serves as the database model schema defining the shape of the `Feedback` table"""
+
+    # Name for the feedback table in the PostgreSQL database
+    __tablename__ = "feedback"
+
+    # Unique ID for the feedback
+    id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
+    # User who submitted the feedback ticket
+    user_id = mapped_column(Integer, ForeignKey("user.id"))
+    user = relationship("UserEntity")
+    # Description of the feedback
+    description: Mapped[str] = mapped_column(String)
+    # Whether the feedback has been resolved or not
+    has_resolved: Mapped[bool] = mapped_column(Boolean)
+
+    @classmethod
+    def from_draft_model(cls, model: Feedback) -> Self:
+        """
+        Class method that converts an `DraftFeedback` model into a `FeedbackEntity`
+
+        Parameters:
+            - model (DraftFeedback): Model to convert into an entity
+        Returns:
+            FeedbackEntity: Entity created from model
+        """
+        return cls(
+            id=model.id,
+            user_id=model.user_id,
+            user=model.user,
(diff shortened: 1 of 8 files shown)


backend/entities/feedback_entity.py
```diff
@@ -0,0 +1,100 @@
...
+    """Serves as the database model schema defining the shape of the `Feedback` table"""
+
+    # Name for the feedback table in the PostgreSQL database
+    __tablename__ = "feedback"
+
+    # Unique ID for the feedback
+    id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
+    # User who submitted the feedback ticket
+    user_id = mapped_column(Integer, ForeignKey("user.id"))
+    user = relationship("UserEntity")
```
ajaygandecha, line 28: Will we need a back-populating field here, such as if students need to see the feedback they have submitted? Regardless, does not seem to be a major concern for now.

backend/entities/feedback_entity.py
```diff
@@ -0,0 +1,100 @@
...
+    # Description of the feedback
+    description: Mapped[str] = mapped_column(String)
+    # Whether the feedback has been resolved or not
+    has_resolved: Mapped[bool] = mapped_column(Boolean)
```
ajaygandecha, line 32: Nit: I think Kris has moved us to a convention where flag fields do not include an `is_` or a `has_` prefix! `resolved` works perfectly here as the field name.

backend/entities/feedback_entity.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @classmethod
+    def from_draft_model(cls, model: Feedback) -> Self:
```
ajaygandecha, line 35: Might be wrong, but I think this method might be redundant due to `from_model` below it! I think these two methods are duplicates, based on their parameters and output.

backend/entities/feedback_entity.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        return cls(
+            id=model.id,
+            user_id=model.user_id,
+            user=model.user,
+            description=model.description,
+            has_resolved=model.has_resolved,
+        )
+
+    def to_model(self, subject: User | None = None) -> Feedback:
```
ajaygandecha, line 70: `subject` might not be needed as an input here!

backend/entities/feedback_entity.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        return Feedback(
+            id=self.id,
+            user_id=self.user_id,
+            user=self.user,
+            description=self.description,
+            has_resolved=self.has_resolved,
+        )
+
+    def to_details_model(self, subject: User | None = None) -> FeedbackDetails:
```
ajaygandecha, line 85: subject might not be needed as an input here!

backend/models/feedback_details.py
```diff
@@ -0,0 +1,14 @@
+from backend.models.feedback import Feedback
+
+__authors__ = ["Matt Vu"]
+__copyright__ = "Copyright 2023"
+__license__ = "MIT"
+
+
+class FeedbackDetails(Feedback):
```
ajaygandecha, line 8: Since there are no back-populating relationship fields, is this model necessary?

backend/models/feedback.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+
+    id: int
+    user_id: int
+    user: User
+    description: str
+    has_resolved: bool
+
+
+class FeedbackForm(BaseModel):
```
ajaygandecha, line 35: This model might not be necessary, since it looks equivalent to `Feedback`!

ajaygandecha (changes requested):
Sorry for the late reply to this @tmattvu! This is looking great, nice work! Left a few comments regarding some of the models, let me know what you think.

KrisJordan:
For frontend purposes, let's get Feedback into its own module, separate from the main module, so that it's not loaded with every page load.

ajaygandecha:
Any updates @tmattvu ?

tmattvu:
> Any updates @tmattvu ?

No sir, but I just finished group reservations :)

ajaygandecha:
Closing branch!

Issue #251, opened 2024-01-30
Add Feedback Buttons/Pages


jadekeegan:
# Add Feedback Buttons/Pages
Add buttons/pages for reporting issues or providing suggestions regarding the CSXL..

## Why?
This will provide users with the opportunity to interact with the CSXL and help us improve the space and site.

## Acceptance Criteria
- [ ] Confirm with Kris how he would like to handle reporting issues/giving feedback - do we want feedback to be internally available, or should we just link to an outside form?
    - [ ] Let @jadekeegan know what Kris decides if not already aware so acceptance criteria can be updated.
- [ ] Create a button to navigate to a suggestions form.
- [ ] ... further acceptance criteria TBA

ajaygandecha:
@KrisJordan, are we still keeping this ticket open, or have we decided to let feedback remain in GitHub and in Slack?

ajaygandecha:
@tmattvu closing this PR until we get #15 working.

PR #257, opened 2024-02-06
Slack Integration and Sidebar Bottom Buttons

krohan03:
This PR will implement the integration of Slack into the CSXL website via a newly created button bar at the bottom of the side navbar. This sidebar also contains a button to the CSXL GitHub codebase and can continue to be populated with other buttons in the future.

### Changelog

- Created a section for buttons at the bottom of the sidebar in navigation component
- Created slack-invite-box widget to contain description and rules of the CSXL slack with a permalink to the CSXL slack
- Created Slack button in bottom bar of navigation component to open slack-invite-box widget as dialog box
- Created GitHub button in bottom bar of navigation component to link to CSXL codebase
- Configured GitHub button to automatically swap between light/dark mode logos based on device setting
- Configured button bar to appear only if user is signed in

### Future Considerations

- Planning to integrate 'Feedback' button from Issue #251 into button bar


frontend/src/app/shared/shared.module.ts
```diff
@@ -36,7 +37,8 @@ import { UserChipList } from './user-chip-list/user-chip-list.widget';
     EventCard,
     EventList,
     UserLookup,
-    UserChipList
+    UserChipList,
+    SlackInviteBox
```
ajaygandecha, line 41: Nit: I might argue that `SlackInviteBox` can be moved to the App Module, since it is only used by one component! In the future though, we probably need to move this into its own module.

frontend/src/app/shared/slack-invite-box/slack-invite-box.widget.html
```diff
@@ -0,0 +1,35 @@
...
+      Experience Lab community?
+    </p>
+    <p>
+      Join the CSXL Slack to get updates about the XL, learn about new
+      opportunities for CS students, or meet your fellow CS majors!
+    </p>
+    <br />
+    <h3>
+      By accepting the below invite and joining the CSXL Slack, you are agreeing
+      to the following:
```
ajaygandecha, line 15: Nit: Might be better if this header has a lower prominence! (like maybe `<h4>` or lower!)

frontend/src/assets/slack-logo.png
ajaygandecha, line 1: This might not be used anymore and can be deleted!

ajaygandecha (changes requested):
Great work @krohan03! Looks great, very excited for this feature. The Slack popup is great too! Just some small nits commented below.

Also, one thing I was noticing is that the images do not necessarily have the best resolution. One potential solution to this is to use icons / SVGs with custom colors applied to ensure it is as sharp as possible and look more like a button! There is a great website called [SimpleIcons](https://simpleicons.org/?q=slack) that provides these. You can download the images in SVG format and then attempt to load them into the HTML. No worries if we do not end up doing this, just a thought! 🤩

frontend/src/app/shared/shared.module.ts
KrisJordan: I concur, may as well go ahead and do this move ahead of initial merge. I'll take a stab.

frontend/src/app/navigation/navigation.component.ts
```diff
@@ -73,6 +75,14 @@ export class NavigationComponent implements OnInit, OnDestroy {
     }
   }
 
+  onSlackInviteClick(): void {
+    const dialogRef = this.slackDialog.open(SlackInviteBox, {
+      width: '1000px',
+      autoFocus: false
```
KrisJordan: I'm going to modify these to be more aligned with latest dialog options and remove the hardcoded width.

frontend/src/app/shared/slack-invite-box/slack-invite-box.widget.html
KrisJordan: I'm going to take a stab at tweaking this language in a push.

frontend/src/assets/slack-logo.png
KrisJordan: I wouldn't be opposed to the color version being the hover state... going to give that a quick shot.

KrisJordan (approved):
I made some changes and am going to go ahead and merge.

I'm going to make a follow-on ticket to be sure we're using the social media shared icons widget rather than the custom button in the sidebar currently. Or, have a strong reason for opting out of it, I believe this will avoid some unnecessary downloads and is a good use case for the shared widget.

Great work here!

KrisJordan:
> Great work @krohan03! Looks great, very excited for this feature. The Slack popup is great too! Just some small nits commented below.
> 
> Also, one thing I was noticing is that the images do not necessarily have the best resolution. One potential solution to this is to use icons / SVGs with custom colors applied to ensure it is as sharp as possible and look more like a button! There is a great website called [SimpleIcons](https://simpleicons.org/?q=slack) that provides these. You can download the images in SVG format and then attempt to load them into the HTML. No worries if we do not end up doing this, just a thought! 🤩

I concur with this, but going to make a separate ticket and go ahead and merge.

krohan03:
> Great work @krohan03! Looks great, very excited for this feature. The Slack popup is great too! Just some small nits commented below.
> 
> Also, one thing I was noticing is that the images do not necessarily have the best resolution. One potential solution to this is to use icons / SVGs with custom colors applied to ensure it is as sharp as possible and look more like a button! There is a great website called [SimpleIcons](https://simpleicons.org/?q=slack) that provides these. You can download the images in SVG format and then attempt to load them into the HTML. No worries if we do not end up doing this, just a thought! 🤩

That makes complete sense, I'll take a look at that and try to get some better resolution icons, thanks!!

PR #274, opened 2024-02-20
[Room Reservations] Query DB instead to get list of reservable rooms

yuvrajjain2003:
## An aside:

This is not a PR to `main` but rather a PR to `room-reservations`. The way Jade and I had decided to move forward with merging strategies for our feature was to send PR into the room reservation branch. Once we have made the necessary changes required to launch the beta version for our feature, we will merge `room-reservations` into main, which will be a giant merge. Was just wondering if you were okay with this strategy @KrisJordan?

## Back to the PR:

Previously, we were just iterating over a hardcoded list to get all the rooms available. This change ensures that we instead query the rooms database to get all the reservable rooms. There is also a test for this method.

However, one thing I did notice is that this query only returned 4 reservable rooms. I am sure that there are more than 4 reservable rooms, so I was wondering how I should go about adding them into the database if anyone was aware? @KrisJordan @ajaygandecha @jadekeegan.

### DIFF
--- a/backend/services/coworking/reservation.py
+++ b/backend/services/coworking/reservation.py
@@ -7,8 +7,7 @@ from typing import Sequence
 from sqlalchemy.orm import Session, joinedload
 from backend.entities.room_entity import RoomEntity
 
-from backend.models.room import Room
-from backend.test.services import room_data
+from backend.models.room_details import RoomDetails
 from ...database import db_session
 from ...models.user import User, UserIdentity
 from ..exceptions import UserPermissionException, ResourceNotFoundException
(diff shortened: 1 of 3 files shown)


backend/services/coworking/reservation.py
```diff
@@ -325,6 +324,31 @@ def _query_confirmed_reservations_by_date(
...
+        Each room entity is converted to a RoomDetails model before being returned.
+
+        Returns:
+            Sequence[RoomDetails]: A sequence of RoomDetails models representing all the reservable rooms, excluding room 'SN156'.
+        """
+
+        rooms = (
+            self._session.query(RoomEntity)
+            .filter(
+                RoomEntity.id != 'SN156',
```
ajaygandecha, line 344: Would probably be a good idea to move `"SN156"` into the coworking policy!

ajaygandecha (commented):
Left one comment here, but will let @KrisJordan review this one ! Great work!

PR #276, opened 2024-02-20
Integrate Event Pagination

jadekeegan:
This PR improves the event page with event pagination in mind and fixes the search bar issues.

## Major Changes
- Updated the date range display to a mat card with arrows, as well as updated the display of the dates.
- Updated the two column layout so the date range display is sticky and the event list is separately scrollable.
- Fixed the search bar query (it needs to be 'en-GB' rather than 'en-US' for the format to be correct for the backend).

### Frontend
All changes were tested thoroughly on the local server. 
1. Without logging in, go to the Events page and click through the paginated events.
2. Ensure that the search bar works properly when a query is added.
3. Log in as a user and repeat steps 1 and 2.


backend/api/events/events.py
```diff
@@ -36,6 +36,49 @@
...
+    """List events in time range via standard backend pagination query parameters for unauthenticated users."""
+
+    pagination_params = EventPaginationParams(
+        order_by=order_by,
+        ascending=ascending,
+        filter=filter,
+        range_start=range_start,
+        range_end=range_end,
+    )
+    return event_service.get_paginated_events(pagination_params)
```
ajaygandecha, line 79: Just curious (because I am not familiar with the pagination code as much), do we need to have a separate `/paginate/unauthenticated` API route, or can we just pass in `subject` into the `/paginate` API and check if the user is unauthenticated that way?

backend/services/event.py
```diff
@@ -55,6 +57,75 @@ def __init__(
...
+
+        if pagination_params.order_by != "":
+            if pagination_params.ascending == "false":
+                statement = statement.order_by(
+                    getattr(EventEntity, pagination_params.order_by).desc()
+                )
+            else:
+                statement = statement.order_by(
+                    getattr(EventEntity, pagination_params.order_by)
+                )
```
ajaygandecha, line 116: Nit: Ternary operator might be able to make this more concise, but not required.

frontend/src/app/event/event-page/event-page.component.ts
```diff
@@ -7,37 +7,67 @@
...
     path: '',
     title: 'Events',
     component: EventPageComponent,
     canActivate: [],
-    resolve: { profile: profileResolver, events: eventResolver }
+    resolve: {
+      profile: profileResolver,
+      page: () =>
+        inject(EventService).list(EventPageComponent.EventPaginationParams)
+    }
```
ajaygandecha, line 65: Since this code is a bit complex, it might be good to include comments / light documentation!

frontend/src/app/event/event-page/event-page.component.ts
```diff
@@ -50,34 +80,97 @@ export class EventPageComponent implements OnInit {
...
+          this.startDate.getTime() > today.setHours(0, 0, 0, 0)
+        ).toString();
+        let paginationParams = this.page.params;
+        paginationParams.range_start = this.startDate.toLocaleString('en-GB');
+        paginationParams.range_end = this.endDate.toLocaleString('en-GB');
+        this.eventService.list(paginationParams).subscribe((page) => {
+          this.eventsPerDay = this.eventService.groupEventsByDate(page.items);
+        });
+      });
+  }
```
ajaygandecha, line 170: Nit: Consider refactoring this into its own function, like `configurePagination`, or instead provide a little bit more documentation (esp in the `routeSubscription` part!)

frontend/src/app/event/event-page/event-page.component.ts
```diff
@@ -91,7 +184,26 @@ export class EventPageComponent implements OnInit {
...
+    } else {
+      paginationParams.range_start = new Date(
+        new Date().setFullYear(new Date().getFullYear() - 100)
+      ).toLocaleString('en-GB');
+      paginationParams.range_end = new Date(
+        new Date().setFullYear(new Date().getFullYear() + 100)
+      ).toLocaleString('en-GB');
+      paginationParams.filter = this.query;
+    }
+    console.log(paginationParams);
```
ajaygandecha, line 202: Leftover log statement!

ajaygandecha (changes requested):
Wow, great job here - AMAZING work on the UI!! Left a few comments, but this is looking very good!

Also a small nit I am adding here rather than in the code, the event list box bottom margin might be higher than the event panel's bottom margin! They are not aligned for me, but it might just be my browser.

backend/api/events/events.py
jadekeegan: Yep it does need a separate route. I'm not sure how it works, but the API route auth vs. not auth takes into account any path that needs authentication I think. So, if there is a subject parameter, it will require authentication.

jadekeegan:
> Is this ready for merge?

Functionality wise yes, but further documentation on Audrey's code has yet to be added.

PR #289, opened 2024-02-26
Add Pydantic models for the office hours feature

sadieamato:
- Added details and base models for OH sections, events, and tickets Note: here, the `OfficeHoursEvent` date field is of type `date` instead of datetime.
- TODO: refactor this field in the OfficeHoursEventEntity as well.
- more info on these data types: https://docs.python.org/3/library/datetime.html


backend/models/office_hours/oh_event.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class OfficeHoursEvent(BaseModel):
+    """
+    Pydantic model to represent an `OfficeHoursEvent` that has not been created yet.
+
+    This model is based on the `OfficeHoursEventEntity` model, which defines the shape
+    of the `OfficeHoursEvent` database in the PostgreSQL database.
+    """
+    id: int | None = None
```
KrisJordan, line 17: I might suggest making this slightly more nuanced so that you can avoid null checks later on. I would recommend:

`OfficeHoursDraft(BaseModel)` - This does not have an `id` field but has everything else you would want to see in a draft version of an event (before it is persisted in the backend).

`OfficeHoursEvent(OfficeHoursDraft)` - This type adds the id field (and any other fields that would only exist in a persisted version, if there are differences).

KrisJordan (changes requested):
Left a note about removing a nullable / optional `id` field in one place. I would generally recommend this strategy anywhere you have `id: int | None`

backend/models/office_hours/oh_event.py
sadieamato: Good point, thank you! With the naming, would you prefer us to use ObjDraft and Obj over using Obj and OBjDetails (where the second name is the one adding on fields)? I saw both used throughout the codebase and was wondering if there were actual use case differences

backend/models/office_hours/oh_event.py
```diff
@@ -0,0 +1,27 @@
...
+    office_hours_section_id: int
+    location_id: int
```
ajaygandecha, line 19: Nit: I might be wrong here so feel free to ignore lol, but it might be good to rename this to `room_id`, since it is a foreign key to the `room` table! That way, later on, it is easier to understand which data / table this ID is referring to

backend/models/office_hours/oh_event.py
```diff
@@ -0,0 +1,27 @@
...
+    type: OfficeHoursType
+    title: str
+    description: str
+    location_description: str
```
ajaygandecha, line 23: Nit: Curious, would this be extra data added to the room? Wondering if this field is necessary to include in addition to `description`! Also, would this potentially be `| None` if no description is supplied, or will the default be ""?

backend/models/office_hours/oh_section.py
```diff
@@ -0,0 +1,16 @@
+from pydantic import BaseModel
+
+__authors__ = ["Sadie Amato"]
+__copyright__ = "Copyright 2024"
+__license__ = "MIT"
+
+
+class OfficeHoursSection(BaseModel):
```
ajaygandecha, line 8: @KrisJordan can't remember, do we have a convention for the names of association entities? Wondering if naming such that "Event" is included in the name would be better since this is the association table btwn `oh__event` and `academics__section`, or if such a name would be too long.

backend/models/office_hours/oh_event.py
KrisJordan: Annoyingly: yes :)

There's actually a sound argument for having three models for the same entity depending on usage:

1. Draft - pre-persistence
2. Plain - Just the model persisted, minimal/no relationships loaded
3. Details - Full model with relationships

The plain models tend to be what's loaded in relationships of a details model (to avoid circularity in details models). For example, ticket details may load info on users via plain models but not user details models (which may load related plain tickets but not details tickets)

ajaygandecha (changes requested):
@sadieamato This is looking great so far! Added a couple of nit comments on thoughts - no real action items!

backend/models/office_hours/oh_event.py
baileymeredith: We were thinking this could be optional extra instructions a professor or TA could add to an event. For example, if students should be sitting in a specific area or if there are special instructions for how to get to the location!

backend/models/office_hours/oh_event.py
baileymeredith: Updated, thank you!

backend/models/office_hours/oh_event.py
baileymeredith: This has been updated now in our recent commit, thank you!!

backend/models/office_hours/oh_section.py
meghansun322: Our office hours section table isn't an association table between `oh__event `and `academics__section`; it's a one-to-many with the `academics__section`. It just looks like one since there are only a few fields!

backend/models/office_hours/oh_event_status.py
```diff
@@ -0,0 +1,26 @@
...
+    num_open: int
+    num_pending: int
+
+
+class StudentOfficeHoursEventStatus(BaseModel):
+    """Pydantic model to represent the status of a given event, including their position in the queues, for a given student."""
+
+    num_open: int
+    num_pending: int
+    position: int
```
ajaygandecha, line 26: Would this ultimately be a field in the ticket model, or would this be queried separately? If queried separately, it might be good to add a reference to a ticket here!

backend/models/office_hours/oh_ticket.py
```diff
@@ -0,0 +1,38 @@
...
+    Pydantic model to represent an `OfficeHoursTicket` that has not been created yet.
+
+    This model is based on the `OfficeHoursTicketEntity` model, which defines the shape
+    of the `OfficeHoursTicket` database in the PostgreSQL database.
+    """
+
+    office_hours_event_id: int
+    description: str
+    type: TicketType
+    state: TicketState
```
ajaygandecha, line 22: Curious, I was unable to find these files! Is this in another branch?

ajaygandecha (commented):
Looking great! Added two comments regarding ticket statuses, otherwise looking super cool!

backend/models/office_hours/oh_ticket.py
meghansun322: The `office-hours` branch will have these enum + entities files, and this was the PR that had initialized it: https://github.com/unc-csxl/csxl.unc.edu/pull/295

We forgot to merge those entities and enums into this current working branch; we just merged those new changes, so those files should exist now on this branch!

backend/models/office_hours/oh_event_status.py
meghansun322: The ticket would be queried separately! This is a model for fetching information about the current status of all tickets that are active/ in the queue for an office hours event.

meghansun322: The `StudentOfficeHoursEventStatus` class represents a model related to an API response we want to make, so it is separate and not related to our ticket model. It's meant to represent a summary of all ticket's status in a queue!

We also didn't want to have a ticket reference here since the information in this model would likely update more often as tickets are being made, called, and closely more frequently when a user-created ticket would be updated. 

This image is what we intended `StudentOfficeHoursEventStatus` class API response model to represent:

backend/models/office_hours/oh_event.py
```diff
@@ -0,0 +1,36 @@
...
+class OfficeHoursEventDraft(BaseModel):
+    """
+    Pydantic model to represent an `OfficeHoursEvent` that has not been created yet.
+
+    This model is based on the `OfficeHoursEventEntity` model, which defines the shape
+    of the `OfficeHoursEvent` database in the PostgreSQL database
+    """
+
+    office_hours_section_id: int
+    room_id: int
```
KrisJordan: I would suggest raising the level of abstraction in the models here, but doing so may mean tweaking the models of `OfficeHoursSection` and `Room` slightly.

The suggested change would be adding a `RoomPartial` model to `backend/models/room.py` and making `nickname` be `| None`: 

~~~
class RoomPartial(Room):
   nickname: str | None = None
~~~ 

Similarly, with `OfficeHoursSection`:

~~~
class OfficeHoursSectionPartial(OfficeHours):
  title: str | None = None
~~~

The reason for moving toward partials (an idea inspired by TypeScript which has a true notion of a Partial version of a type where fields can be null) is so the client can pass the full object if it'd like and not have to reshape the request to form a `OfficeHoursEventDraft` object to extract `id`s of foreign key relationships (which is hopefully a lower level abstraction, even though it winds up getting passed in anyway). 

This way, from the client-side, you'll be able to make requests like:

~~~
this.httpClient.post('/office-hours/event', {
   office_hours_section: officeHoursSection,
   room: room,
   type: type,
   /* ... */
});
~~~

Notice the slight nicety that on the client side you can think in terms of the objects you otherwise have in the system (like the officeHoursSection object and room object) rather than having to remember to pull out their IDs each time.

The primary motivation here is trying to give the client the ability to think less about the database structure (where do I have to pull out an `id` in my request vs. just give an object) and more about the objects in the system.

Finally, this helps avoid accidental errors such as choosing the wrong type for an `id`. The Room type's `id` is actually a `str`!

backend/models/office_hours/oh_event.py
KrisJordan, line 1: I would perhaps remove the prefixes of these file names. Totally fine with having the model names prefixed with `OfficeHours`, though.

backend/models/office_hours/oh_event.py
```diff
@@ -0,0 +1,36 @@
+from pydantic import BaseModel
+from datetime import datetime, date
+from oh_type import OfficeHoursType
```
KrisJordan: Does this need to be a relative import? `.oh_type`

backend/models/office_hours/oh_event_status.py
```diff
@@ -0,0 +1,26 @@
...
+__copyright__ = "Copyright 2024"
+__license__ = "MIT"
+
+
+class TAOfficeHoursEventStatus(BaseModel):
+    """Pydantic model to represent the status of a given event, including tickets in the queue, for a given TA."""
+
+    active_tickets: Sequence[OfficeHoursTicket]
+    num_open: int
+    num_pending: int
```
KrisJordan, line 18: Can we iterate on names of these fields? I'd like to avoid abbreviations, if possible. Some alternative ideas:

`open`
`open_count`
`open_tickets_count`

Additionally, thoughts on `queued` versus `pending`?  The same language should propagate to `ticket_state`. There, `pending` is the only `ing` state. 

Going for a super explicit name here has my vote, I think.

backend/models/office_hours/oh_section.py
```diff
@@ -0,0 +1,27 @@
...
+
+class OfficeHoursSection(OfficeHoursSectionDraft):
+    """
+    Pydantic model to represent an `OfficeHoursSection` that has not been created yet.
+
+    This model is based on the `OfficeHoursSectionEntity` model, which defines the shape
+    of the `OfficeHoursSection` database in the PostgreSQL database.
+    """
+
+    id: int | None = None
```
KrisJordan, line 27: Make `id` just `int` at this point, per reasons above. Once persisted, it will always have an `id`.

backend/models/office_hours/oh_section_details.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    Pydantic model to represent an `OfficeHoursSection`, including back-populated
+    relationship fields.
+
+    This model is based on the `OfficeHoursSectionEntity` model, which defines the shape
+    of the `OfficeHoursSection` database in the PostgreSQL database.
+    """
+
+    sections: list[Section]
+    events: list[OfficeHoursEvent]
```
KrisJordan, line 20: It seems like there may need to be another model for getting upcoming events to drive an office hours section home page as opposed to all events in the history of a section.

backend/models/office_hours/oh_ticket.py
```diff
@@ -0,0 +1,38 @@
...
+
+class OfficeHoursTicket(OfficeHoursTicketDraft):
+    """
+    Pydantic model to represent an `OfficeHoursTicket`.
+
+    This model is based on the `OfficeHoursTicketEntity` model, which defines the shape
+    of the `OfficeHoursTicket` database in the PostgreSQL database.
+    """
+
+    id: int | None = None
```
KrisJordan, line 35: Another place where `id` will always be an `int` and never `None`.

backend/models/office_hours/oh_ticket.py
```diff
@@ -0,0 +1,38 @@
...
+
+class OfficeHoursTicketDraft(BaseModel):
+    """
+    Pydantic model to represent an `OfficeHoursTicket` that has not been created yet.
+
+    This model is based on the `OfficeHoursTicketEntity` model, which defines the shape
+    of the `OfficeHoursTicket` database in the PostgreSQL database.
+    """
+
+    office_hours_event_id: int
```
KrisJordan: If we go down the road of slightly raising the level of abstraction, my vote here would be for this to be `office_hours_event: OfficeHoursEventPartial`

backend/models/office_hours/oh_ticket.py
KrisJordan, line 22: `state` is not needed at the point of draft, yeah?

backend/models/office_hours/oh_ticket.py
```diff
@@ -0,0 +1,38 @@
...
+class OfficeHoursTicket(OfficeHoursTicketDraft):
+    """
+    Pydantic model to represent an `OfficeHoursTicket`.
+
+    This model is based on the `OfficeHoursTicketEntity` model, which defines the shape
+    of the `OfficeHoursTicket` database in the PostgreSQL database.
+    """
+
+    id: int | None = None
+    caller_id: int | None = None
```
KrisJordan, line 36: I would go ahead and make `caller` be of type `PublicUser | None` https://github.com/unc-csxl/csxl.unc.edu/blob/feature/oh-backend-api-models/backend/models/public_user.py or `SectionMember` (not seeing that model, but isn't that just a simple join table-style relationship back to a user?)

backend/models/office_hours/oh_ticket_details.py
```diff
@@ -0,0 +1,21 @@
...
+    Pydantic model to represent an `OfficeHoursSection`, including back-populated
+    relationship fields.
+
+    This model is based on the `OfficeHoursSectionEntity` model, which defines the shape
+    of the `OfficeHoursSection` database in the PostgreSQL database.
+    """
+
+    event: OfficeHoursEvent
+    creators: list[SectionMember]
+    caller: SectionMember | None = None
```
KrisJordan, line 21: It seems these two fields should be elevated to `OfficeHoursTicket`.

KrisJordan (commented):
Seeing some great progress here; made a number of notes to consider trying to think ahead toward front-end concerns and what the experience will be in consuming the API. Some of these changes are slightly more effortful to implement from backend side, but will make the client-side more pleasant.

backend/models/office_hours/oh_ticket.py
meghansun322: @KrisJordan We have our `caller` field a part the `OfficeHoursTicketDetails` model and then caller id is the standard ticket model.  We saw your comment about elevating both creators and caller from `OfficeHoursTicketDetails` to `OfficeHoursTicket`. Both fields are relationship fields in our ticket model, and we saw the ticket details model as a way to represent a model's full relationships. 

Would having the caller relationship field in just the `OfficeHoursTicket` still be better from a design or usability standpoint? All endpoints related to returning information about the ticket use the `OfficeHoursTicketDetails`

In addition, is there a benefit of using PublicUser over SectionMember? As if right now, we have it so that the caller and creators are mapped to `SectionMember` -- this enforces so that users involved in office hours related activities would have need to be members of the course section!

backend/models/office_hours/oh_event.py
baileymeredith: This has been adjusted now, thank you!!

backend/models/office_hours/oh_ticket.py
baileymeredith: This has been updated, thank you!

PR #262, opened 2024-02-09
Refactored gear, organization form, and role creation

bwgoulet:
This PR is a refactor of a few of the XL's features - namely, editable forms that were previously attached to the pencil icon are now accessible via gears. Additionally, the organization form has been updated for readability, and new roles are automatically created whenever a new organization is added.

## Changelog
- Refactored editable forms to be accessed by gear icon
- Updated the organization form for readability 
- Associated Roles are now created upon organization creation

PR #294, opened 2024-02-27
241 refactor organization management

maxgoetz:
This PR updates organization editing to have a full page devoted to organization and event admins.
**Changes for include:**
1. Moving the organization and event admin pages from within the user admin tab to its own dedicated space, accessed by the gear tool added to the organization and event pages.
3. Giving all users with organization admin permissions the ability to access the organization and event admin page, but ensuring that they only have access to view and edit their own organizations.
4. Removing the edit tool within an organization's details.
5. Ensuring that only a root user can create and delete organizations through dedicated buttons that only appear for the root user.


maxgoetz:
This PR is no longer needed. All code is now in PR #262.

PR #295, opened 2024-02-27
281 Create Office Hours Entities

meghansun322:
Table and Entities for Office Hours. 

https://miro.com/app/board/uXjVNwL9kIo=/

### DIFF
--- a/backend/entities/office_hours/oh_event_entity.py
+++ b/backend/entities/office_hours/oh_event_entity.py
@@ -1,7 +1,7 @@
 """Definition of SQLAlchemy table-backed object mapping entity for Office Hour Sections."""
 
-from datetime import datetime
-from sqlalchemy import DateTime, ForeignKey, Integer, String
+from datetime import datetime, date
+from sqlalchemy import Date, DateTime, ForeignKey, Integer, String
 from sqlalchemy.orm import Mapped, mapped_column, relationship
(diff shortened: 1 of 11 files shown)


backend/entities/office_hours/oh_event_entity.py
```diff
@@ -0,0 +1,59 @@
...
+        SQLAlchemyEnum(OfficeHoursType), nullable=False
+    )
+    # Name of event
+    title: Mapped[str] = mapped_column(String, nullable=False)
+    # Description of event
+    description: Mapped[str] = mapped_column(String, nullable=True)
+    # Description of the location; allows for instructors to write note about attending office hours
+    location_description: Mapped[str] = mapped_column(String, nullable=True)
+    # Date of the event
+    date: Mapped[datetime] = mapped_column(DateTime, nullable=False)
```
sadieamato, line 38: Thinking more on this field, I think we should use the `date` class instead of the datetime class since we have a start and end time as separate fields. See info on the classes [here](https://docs.python.org/3/library/datetime.html)
The line could look like this: `date: Mapped[date] = mapped_column(Date, nullable=False)`

sadieamato (changes requested):
LGTM, aside from the one comment i left about changing the date field to be a `date` type!

PR #296, opened 2024-02-27
Community Agreement Bug Fix

tmattvu:
- Few users are currently experiencing UI bug where their browser's interface (particularly some Safari layouts) is blocking the accept button on the community agreement pop up
- Added spacing on bottom of button that should allow user to scroll a little bit more
- In case, also added:  padding-bottom: env(safe-area-inset-bottom);
    - This should ensure that the browsers interface or device's physical features should not block anything on the pop up


KrisJordan (approved):
Lgtm

PR #298, opened 2024-02-28
Fix fields in OH entities to match models

maddyandrews:
This PR makes changes to our entities that reflect the modifications requested of our Pydantic models.

- Several string fields are now non-nullable, with a default of empty string, if they are optional fields
- `location_id` is changed to `room_id`
- Defaults given for fields with DateTime types

### DIFF
--- a/backend/entities/office_hours/oh_event_entity.py
+++ b/backend/entities/office_hours/oh_event_entity.py
@@ -8,7 +8,6 @@ from backend.models.office_hours.oh_type import OfficeHoursType
 
 
 from ..entity_base import EntityBase
-from typing import Self
 from sqlalchemy import Enum as SQLAlchemyEnum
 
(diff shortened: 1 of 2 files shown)


backend/entities/office_hours/oh_event_entity.py
```diff
@@ -28,12 +27,12 @@ class OfficeHoursEventEntity(EntityBase):
     type: Mapped[OfficeHoursType] = mapped_column(
         SQLAlchemyEnum(OfficeHoursType), nullable=False
     )
-    # Name of event
-    title: Mapped[str] = mapped_column(String, nullable=False)
```
maddyandrews, line 32: we realized we don't actually need a title for events, since we plan on using the event type as the label

meghansun322:
LGTM!!

PR #300, opened 2024-02-29
279 add oheventservice  corresponding methods

baileymeredith:
Branch title is misleading... we initially wanted to start on services and decided to complete the to_model and from_model methods in all our entity classes first. 
- edited to_details_model in section_member_entity and room_entity to reflection relations
- created section_member_details model to reflect relations for ticket information
- fixed import statements in various model files

### DIFF
--- a/backend/entities/office_hours/oh_event_entity.py
+++ b/backend/entities/office_hours/oh_event_entity.py
@@ -1,8 +1,11 @@
 """Definition of SQLAlchemy table-backed object mapping entity for Office Hour Sections."""
 
 from datetime import datetime, date
+from typing import Self
 from sqlalchemy import Date, DateTime, ForeignKey, Integer, String
 from sqlalchemy.orm import Mapped, mapped_column, relationship
+from backend.models.office_hours.oh_event import OfficeHoursEvent
+from backend.models.office_hours.oh_event_details import OfficeHoursEventDetails
 
 from backend.models.office_hours.oh_type import OfficeHoursType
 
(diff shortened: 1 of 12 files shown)


backend/entities/office_hours/oh_event_entity.py
```diff
@@ -50,10 +52,72 @@ class OfficeHoursEventEntity(EntityBase):
...
+            OfficeHoursEventEntity: Entity created from model
+        """
+        return cls(
+            id=model.id,
+            office_hours_section_id=model.office_hours_section_id,
+            room_id=model.room_id,
+            type=model.type,
+            description=model.description,
+            location_description=model.location_description,
+            date=model.event_date,
```
meghansun322, line 79: I believe this should be `date=model.date` since the field that represents `date` in the `OfficeHoursEvent` class is named that! I only discovered this when trying to add test data for this table!

meghansun322: Maybe we could just go back to naming it from `event_date` to `date`. I think we wanted to name it `event_date` at first since we wanted to avoid confusion with the `date` type

meghansun322: Or we can change all `date` fields in our models to `event_date`

backend/entities/office_hours/oh_ticket_entity.py
```diff
@@ -58,9 +62,75 @@ class OfficeHoursTicketEntity(EntityBase):
...
+            - model (OfficeHoursTicket): Model to convert into an entity
+        Returns:
+            OfficeHoursTicketEntity: Entity created from model
+        """
+        return cls(
+            id=model.id,
+            caller_id=model.caller_id,
+            have_concerns=model.have_concerns,
+            caller_notes=model.caller_notes,
+            office_hours_event_id=model.office_hours_event_id,
```
meghansun322, line 87: We have a naming difference here as well. I think on the entity side, our field is named `event_id` while on the model side, we have it name `office_hours_event_id`. Maybe we should should consider having them named the same?

But with that, I think this line would be `event_id=model.office_hours_event_id`

Only discovered this when I was trying to put test data into the tables!

backend/entities/office_hours/oh_ticket_entity.py
```diff
@@ -58,9 +62,75 @@ class OfficeHoursTicketEntity(EntityBase):
...
+
+        Returns:
+            OfficeHoursTicket: `OfficeHoursTicket` object from the entity
+        """
+        return OfficeHoursTicket(
+            id=self.id,
+            caller_id=self.caller_id,
+            have_concerns=self.have_concerns,
+            caller_notes=self.caller_notes,
+            office_hours_event_id=self.office_hours_event_id,
```
meghansun322, line 107: To follow up the previous, then it would be `self.event_id` instead!

backend/entities/office_hours/oh_event_entity.py
baileymeredith: These were all great catches, thank you!! This has been updated now from event_date to date.

backend/entities/office_hours/oh_ticket_entity.py
baileymeredith: I changed the naming in both oh_ticket_entity.py and oh_ticket.py for it to be oh_event_id and oh_event; this helps avoid confusion with the separate event entity and is more standardized with the rest of our naming conventions.

backend/entities/office_hours/oh_ticket_entity.py
baileymeredith: This has been updated, thank you!! (changed to oh_event_id=self.oh_event_id)

backend/entities/office_hours/oh_event_entity.py
```diff
@@ -50,10 +52,72 @@ class OfficeHoursEventEntity(EntityBase):
...
-    location_id: Mapped[int] = mapped_column(ForeignKey("room.id"), nullable=False)
-    location: Mapped["RoomEntity"] = relationship("RoomEntity")
+    room_id: Mapped[int] = mapped_column(ForeignKey("room.id"), nullable=False)
+    room: Mapped["RoomEntity"] = relationship("RoomEntity")
 
     # NOTE: One-to-many relationship of OfficeHoursEvent to tickets
     tickets: Mapped[list["OfficeHoursTicketEntity"]] = relationship(
-        back_populates="event", cascade="all, delete"
-    )
+        back_populates="event", cascade="all, delete")
```
meghansun322, line 60: One quick, small thing! Since we just changed the name of the event field in `OfficeHoursTicketEntity` from `event` to `oh_event`, we probably need to change this back populates reference to `oh_event`! 

Looks great otherwise!!

baileymeredith: This has been updated now, thanks!!

PR #307, opened 2024-03-01
Added api routes for events + services, Added skeleton service code

baileymeredith:
Add OH section API endpoints with new models:

- Created api routes for OfficeHoursSections (get, create, update, delete)

Add OH event API endpoints with new models:

- Created api routes for OfficeHoursEvents
    - get by section id, get upcoming events by section id, get upcoming events by user, create, update, delete
    - added fix User subject to api/office_hours/oh_section.py

Add Service files skeleton — func defs + starter thoughts

- all function bodies are WIP/TODO
- changed `update` section & event API routes to take in non-draft model, so that the service can use an id from the db
- OHSectionService: create, delete, get sections by term, get user sections by term, update
- OHEventService: create, update, delete, get events by section, get upcoming events by section, get upcoming events for a user

### DIFF
--- a/backend/api/office_hours/oh_event.py
+++ b/backend/api/office_hours/oh_event.py
@@ -4,6 +4,8 @@ This API is used to access OH Event data."""
 
 from datetime import datetime, timedelta
 from fastapi import APIRouter, Depends
+
+from ...models.office_hours.oh_ticket_details import OfficeHoursTicketDetails
 from ...models.coworking.time_range import TimeRange
 from ...models.office_hours.oh_event import OfficeHoursEvent, OfficeHoursEventDraft
 from ...models.office_hours.oh_event_details import OfficeHoursEventDetails
@@ -40,7 +42,9 @@ def new_oh_event(
     return oh_event_service.create(subject, oh_event)
 
 
-@api.put("/{oh_event_id}", response_model=OfficeHoursEventDetails, tags=["Office Hours"])
+@api.put(
+    "/{oh_event_id}", response_model=OfficeHoursEventDetails, tags=["Office Hours"]
+)
 def update_oh_event(
     oh_event: OfficeHoursEvent,
     subject: User = Depends(registered_user),
@@ -67,9 +71,14 @@ def delete_oh_event(
     oh_event: OfficeHoursEventDetails = oh_event_service.get_event_by_id(oh_event_id)
     return oh_event_service.delete(subject, oh_event)
 
-@api.get("/{oh_event_id}", response_model=OfficeHoursEventDetails, tags=["Office Hours"])
-def get_oh_section_by_id(
-    oh_event_id: int, subject: User = Depends(registered_user), oh_event_service: OfficeHoursEventService = Depends()
+
+@api.get(
+    "/{oh_event_id}", response_model=OfficeHoursEventDetails, tags=["Office Hours"]
+)
+def get_oh_event_by_id(
+    oh_event_id: int,
+    subject: User = Depends(registered_user),
+    oh_event_service: OfficeHoursEventService = Depends(),
 ) -> OfficeHoursEventDetails:
     """
     Gets an OH event by OH event ID
@@ -80,7 +89,9 @@ def get_oh_section_by_id(
     return oh_event_service.get_event_by_id(subject, oh_event_id)
 
 
-@api.get("/upcoming", response_model=list[OfficeHoursEventDetails], tags=["Office Hours"])
+@api.get(
+    "/upcoming", response_model=list[OfficeHoursEventDetails], tags=["Office Hours"]
+)
 def get_upcoming_oh_events_by_user(
     start: datetime = datetime.now(),
     subject: User = Depends(registered_user),
@@ -95,3 +106,23 @@ def get_upcoming_oh_events_by_user(
     """
     time_range = TimeRange(start=start, end=end)
     return oh_event_service.get_upcoming_events_by_user(subject, time_range)
+
+
+@api.get(
+    "/{oh_event_id}/tickets",
+    response_model=list[OfficeHoursTicketDetails],
+    tags=["Office Hours"],
+)
+def get_oh_tickets_by_event(
+    oh_event_id: int,
+    subject: User = Depends(registered_user),
+    oh_event_service: OfficeHoursEventService = Depends(),
+) -> list[OfficeHoursTicketDetails]:
+    """
+    Gets list of OH tickets by OH event
+
+    Returns:
+        list[OfficeHoursTicketDetails]: OH tickets within the given event
+    """
+    oh_event: OfficeHoursEventDetails = oh_event_service.get_event_by_id(oh_event_id)
+    return oh_event_service.get_event_tickets(subject, oh_event)
--- a/backend/api/office_hours/oh_event.py
+++ b/backend/api/office_hours/oh_event.py
@@ -5,7 +5,7 @@ This API is used to access OH Event data."""
 from datetime import datetime, timedelta
 from fastapi import APIRouter, Depends
 from backend.models.coworking.time_range import TimeRange
-from backend.models.office_hours.oh_event import OfficeHoursEventDraft
+from backend.models.office_hours.oh_event import OfficeHoursEvent, OfficeHoursEventDraft
 from backend.models.office_hours.oh_event_details import OfficeHoursEventDetails
 from backend.services.office_hours.oh_event import OfficeHoursEventService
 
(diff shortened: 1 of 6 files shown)


backend/services/office_hours/oh_event.py
```diff
@@ -0,0 +1,111 @@
...
+        """Deletes an office hours event.
+
+        Args:
+            subject: a valid User model representing the currently logged in User
+            oh_event_id: ID of office hours event to delete
+        """
+        #TODO
+
+    
+    def get_events_by_section(self, subject: User, oh_section_id: int) -> list[OfficeHoursEventDetails]:
```
maddyandrews, line 72: would this make more sense in the OH section service? it might be more efficient to query office hours sections and return the list of their events (rather than query events and return all that match the section id). i think this is fine too though!!

backend/api/office_hours/oh_section.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    Adds a new OH section to the database
+
+    Returns:
+        OfficeHoursSectionDetails: OH Section created
+    """
+    return oh_section_service.create(subject, oh_section)
+
+
+@api.get("/{term_id}", response_model=list[OfficeHoursSectionDetails], tags=["Office Hours"])
```
meghansun322, line 40: If we call this endpoint, then our api would be something like `"/office-hours/section/{term_id}"`. I believe this is typically the convention if we wanted to get an office hours section by an id -- this path could be misleading as I would think the id in the last part of the url would represent a section id rather than term id.

Since we are trying to query from a category such as term, maybe we could add a term before the term id so it's more apparent that the end point if querying by term id and not a section id!

Maybe something like this, but we could discuss more about how it should look!
`"/office-hours/section/term/{term_id}"`

backend/api/office_hours/oh_section.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    Gets list of OH sections by term ID
+
+    Returns:
+        list[OfficeHoursSectionDetails]: OH sections within the given term
+    """
+    return oh_section_service.get_sections_by_term(subject, term_id)
+
+
+@api.get("", response_model=list[OfficeHoursSectionDetails], tags=["Office Hours"])
```
meghansun322, line 53: Similar to the comment above about potentially specifying term as a specific category within our api path! Also, If we are trying to get Office Hours Section info based on a term id, we may need to have the id a part of the endpoint string. 

`"term/{term_id}"`

Let me know what you guys think!!

backend/api/office_hours/oh_section.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    Gets list of OH sections the currrent user is in during a given term
+
+    Returns:
+        list[OfficeHoursSectionDetails]: User's OH sections within the given term
+    """
+    return oh_section_service.get_user_sections_by_term(subject, term_id)
+
+
+@api.put("", response_model=OfficeHoursSectionDetails, tags=["Academics"])
```
meghansun322, line 68: Should be "Office Hours"?

backend/api/office_hours/oh_section.py
```diff
@@ -0,0 +1,92 @@
...
+__copyright__ = "Copyright 2024"
+__license__ = "MIT"
+
+
+api = APIRouter(prefix="/api/office-hours/section")
+openapi_tags = {
+    "name": "Office Hours",
+    "description": "Office hours section functionality",
+}
+
```
meghansun322, line 23: Similar to Office Hours Events, do you think we should have an endpoint to get an office hours section by `id`? Since we have established relationships in the Office Hours Section model/details, having this model could be helpful to display the related in the frontend!

meghansun322 (commented):
This looks sooo good so far!! Left a few small comment :) Thank you guys for doing this 🫶

backend/api/office_hours/oh_event.py
```diff
@@ -0,0 +1,118 @@
...
+__license__ = "MIT"
+
+
+api = APIRouter(prefix="/api/office-hours/event")
+openapi_tags = {
+    "name": "Office Hours",
+    "description": "Office hours event functionality",
+}
+
+
```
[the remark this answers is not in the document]
maddyandrews: i think this would be a good idea too!

backend/api/office_hours/oh_event.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    Gets list of upcoming OH events within a time range by section ID
+
+    Returns:
+        list[OfficeHoursSectionDetails]: OH events associated with a given section in a time range
+    """
+    time_range = TimeRange(start=start, end=end)
+    return oh_event_service.get_upcoming_events_by_section(
+        subject, oh_section_id, time_range
+    )
```
KrisJordan, line 101: I wonder if this service method and the previous one can be condensed to a single service method? This feels like a specialization of the previous.

backend/api/office_hours/oh_event.py
```diff
@@ -0,0 +1,118 @@
...
+    subject: User = Depends(registered_user),
+    oh_event_service: OfficeHoursEventService = Depends(),
+):
+    """
+    Deletes an OfficeHoursEvent from the database
+    """
+    return oh_event_service.delete(subject, oh_event_id)
+
+
+@api.get("/{section_id}", response_model=list[OfficeHoursEventDetails], tags=["Office Hours"])
```
KrisJordan, line 69: Should this be moved to the section API? Even though is lists events, it feels logically like a child route of a section.

backend/api/office_hours/oh_section.py
KrisJordan: Agree with this concern. Could also make the route feel like a child of a term route if willing to modify academics

KrisJordan (commented):
Looking good! Left a few minor suggestions along the way.

backend/api/office_hours/oh_event.py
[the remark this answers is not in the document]
baileymeredith: I've added this now, I also think it's a great idea to have, thank you!

backend/api/office_hours/oh_section.py
baileymeredith: I've added this in the recent commits, thank you!

backend/api/office_hours/oh_section.py
baileymeredith: This is a good point, thank you! I've changed the endpoint to be "/office-hours/section/{term_id}" and chose to not modify academics (though we can in the future if we want to change that direction).

backend/api/office_hours/oh_section.py
baileymeredith: I've updated this, thank you!!

backend/api/office_hours/oh_event.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Returns:
+        OfficeHoursEventDetails: OH Event created
+    """
+    return oh_event_service.create(subject, oh_event)
+
+
+@api.put("", response_model=OfficeHoursEventDetails, tags=["Academics"])
+def update_oh_event(
+    oh_event: OfficeHoursEventDraft,
```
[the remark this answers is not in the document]
baileymeredith: I've updated this now to be an OfficeHoursEvent!

backend/api/office_hours/oh_event.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    Adds a new OH event to the database
+
+    Returns:
+        OfficeHoursEventDetails: OH Event created
+    """
+    return oh_event_service.create(subject, oh_event)
+
+
+@api.put("", response_model=OfficeHoursEventDetails, tags=["Academics"])
```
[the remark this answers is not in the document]
baileymeredith: I've updated this in all edit paths, thank you!!

backend/api/office_hours/oh_event.py
```diff
@@ -0,0 +1,118 @@
...
+@api.delete("/{oh_event_id}", response_model=None, tags=["Office Hours"])
+def delete_oh_event(
+    oh_event_id: int,
+    subject: User = Depends(registered_user),
+    oh_event_service: OfficeHoursEventService = Depends(),
+):
+    """
+    Deletes an OfficeHoursEvent from the database
+    """
+    return oh_event_service.delete(subject, oh_event_id)
```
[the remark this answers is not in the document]
baileymeredith: This is good to know, thank you! I've updated this in all places applicable.

backend/api/office_hours/oh_event.py
baileymeredith: Great point, I've moved this into the section API!

backend/api/office_hours/oh_event.py
baileymeredith: I've combined these two and also have moved them into the section service file + section api file (as I think this makes more sense to retrieve events in the section files like you mentioned in a previous comment). Right now we have time_range as an optional parameter in the service method.

PR #308, opened 2024-03-01
303 add ticket api and service skeleton

sadieamato:
### Add OH ticket API and Service skeleton

Add OH ticket API endpoints with new models

- create, update (for students editing), update state, get by section id, get by section id and user, get by event

Add Service files skeleton — func defs + starter thoughts

- all function bodies are WIP/TODO
- changed `update` ticket API route to take in non-draft model, so that the service can use an id from the db
- OHTicketService: create, get tickets by section, get tickets by section and user, get tickets by event, update, & update state

### DIFF
--- a/backend/api/office_hours/oh_ticket.py
+++ b/backend/api/office_hours/oh_ticket.py
@@ -9,6 +9,7 @@ from backend.models.office_hours.oh_ticket import (
     OfficeHoursTicketDraft,
 )
 from backend.models.office_hours.oh_ticket_details import OfficeHoursTicketDetails
+from backend.services.office_hours.oh_ticket import OfficeHoursTicketService
 from ..authentication import registered_user
 from ...models import User
 
@@ -41,12 +42,31 @@ def new_oh_ticket(
 
 
 @api.get(
-    "/{section_id}",
+    "/{oh_ticket_id}",
+    response_model=OfficeHoursTicketDetails,
+    tags=["Office Hours"],
+)
+def get_oh_ticket_by_id(
+    oh_ticket_id: int,
+    subject: User = Depends(registered_user),
+    oh_ticket_service: OfficeHoursTicketService = Depends(),
+) -> OfficeHoursTicketDetails:
+    """
+    Gets an OH ticket by its id
+
+    Returns:
+        OfficeHoursTicketDetails: OH ticket with the given id
+    """
+    return oh_ticket_service.get_ticket_by_id(subject, oh_ticket_id)
+
+
+@api.get(
+    "/{oh_section_id}",
     response_model=list[OfficeHoursTicketDetails],
     tags=["Office Hours"],
 )
 def get_oh_tickets_by_section(
-    section_id: int,
+    oh_section_id: int,
     subject: User = Depends(registered_user),
     oh_ticket_service: OfficeHoursTicketService = Depends(),
 ) -> list[OfficeHoursTicketDetails]:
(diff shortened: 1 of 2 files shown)


backend/api/office_hours/oh_ticket.py
```diff
@@ -0,0 +1,123 @@
...
+    return oh_ticket_service.create(subject, oh_ticket)
+
+
+@api.get(
+    "/{section_id}",
+    response_model=list[OfficeHoursTicketDetails],
+    tags=["Office Hours"],
+)
+def get_oh_tickets_by_section(
+    section_id: int,
```
meghansun322, line 49: Just a quick thought! Since we are working with both an academic section and an office hours section, maybe we could specify the variable name to be `oh_section_id` to hopefully avoid confusion?

backend/api/office_hours/oh_ticket.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    Gets list of OH tickets by OH event
+
+    Returns:
+        list[OfficeHoursTicketDetails]: OH tickets within the given event
+    """
+    return oh_ticket_service.get_tickets_by_event(subject, event_id)
+
+
+@api.put("", response_model=OfficeHoursTicketDetails, tags=["Academics"])
```
meghansun322, line 95: Should be "Office Hours" tag?

backend/api/office_hours/oh_ticket.py
```diff
@@ -0,0 +1,123 @@
...
+    Updates an OfficeHoursTicket to the database
+
+    Returns:
+        OfficeHoursTicketDetails: OH Ticket updated
+    """
+    return oh_ticket_service.update(subject, oh_ticket)
+
+
+# TODO: Please check this api route
+@api.put("/state", response_model=OfficeHoursTicketDetails, tags=["Academics"])
```
meghansun322, line 111: Same here! Just a small change :)

backend/api/office_hours/oh_ticket.py
```diff
@@ -0,0 +1,123 @@
...
+__license__ = "MIT"
+
+
+api = APIRouter(prefix="/api/office-hours/ticket")
+openapi_tags = {
+    "name": "Office Hours",
+    "description": "Office hours ticket functionality",
+}
+
+
```
meghansun322, line 27: Do you guys think we would also need an API route to get a ticket by `id`? I guess the post and update returns the Ticket Details model, but it may help to have this endpoint still? If not explicitly used on the front end, it could still be a helpful endpoint for development purposes. Let me know what you guys think!

meghansun322 (commented):
Yayyy this is looking amazing!! Left a few comments but nothing major 😎

backend/api/office_hours/oh_ticket.py
sadieamato: Added this in my most recent commit! Yes, i think it will be helpful for us to have even if there's not a use case in actual usage.

backend/api/office_hours/oh_ticket.py
sadieamato: Fixed!

backend/api/office_hours/oh_ticket.py
sadieamato: Fixed!

backend/api/office_hours/oh_ticket.py
sadieamato: Renamed this in recent commit! Plus changed another place where we just wrote `event_id` to be `oh_event_id`

PR #310, opened 2024-03-02
Make User Lookup widget more generic

KrisJordan:
The User Lookup widget has a few characteristics that are specific to its initial use in the EventEditorComponent. This PR makes the component more generic and refactors a few minor improvements:

1. Make the control label an input of the control
2. Use a notion of the control being "disabled" rather than have it based on some "adminPermission" which made sense in the context of EventEditorComponent
3. Add an optional `maxSelected` input property which is `null` by default. If set to a number, the control prevents the addition of more users. (My new use case for this widget elsewhere is a select one.)

Since this involved refactoring some of the `event-editor.component.html` file, I made some small improvements there, as well.

### DIFF
--- a/frontend/src/app/event/event-editor/event-editor.component.html
+++ b/frontend/src/app/event/event-editor/event-editor.component.html
@@ -2,10 +2,7 @@
 <form
   [formGroup]="eventForm"
   (ngSubmit)="onSubmit()"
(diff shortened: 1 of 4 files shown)


jadekeegan (approved):
LGTM!

Issue #288, opened 2024-02-25
Restrict reservations when XL is closed.


yuvrajjain2003:
## Why?

There is a current problem with the model of room reservations because Sally Student can virtually book out a room anytime in a week. This is 'false advertisement' because the XL is actually closed during the weekend and during university holidays. We need some way of interacting with `operating_hours` to restrict reservations during these times.

## Acceptance Criteria

- [x] Students should not be allowed to make room reservations for time slots during weekends.
- [x] Students should not be allowed to make room reservations when it is a university holiday for example Wellness days.
- [x] The frontend table should not allow reservations for these time slots.

PR #311, opened 2024-03-02
Dynamic Room Reservation Table based on Operating hours

### LINKED ISSUE #264: Dynamic opening hours based on the operating hours table
# Why?

Currently, the room reservation table has hard-coded values to allow users to make reservation from 10 am to 6 pm, but we would like to modify this so that it is relative to the operating hours. If the operating hours were to change in the future, then we would not need to modify these hard coded values. Also, it helps with showing when the XL is closed like during Wellness days. 

# Acceptance Criteria

- [x] Replace hard coded values in the backend for drafting new room reservation with relative operating hour times.
- [x] Only allow users to reserve rooms in the frontend when the XL would be open. This means, that users should not be allowed to reserve rooms for the weekend or during campus holidays like the Wellness days. 
- [x] Replace the hard coded values in the frontend for the reservation times with relative timeslots for operating hours that will be requested from the backend.

yuvrajjain2003:
closes #264, #288 

This PR addresses the issue of students being able to reserve rooms even when the XL is closed, which served as "false advertising." Now Sally Student should only be able to reserve rooms based on the operating hours of the XL. That is, Sally can only reserve a room on a date when the XL has listed operating hours. Addressing this issue in turn required us to make the table dynamic so that it changes size based on the duration of the operating hours.

## Changes in the Backend

- Instead of hardcoding values to only allow reservations from 10 am to 6 pm, we now query the database to retrieve the list of operating hours. This allows our table logic to be more abstract.
- We introduced `ReservedMapDetails` as a pydantic model to facilitate this data transfer to the frontend. 
- We added some helper functions to allow this dynamic calculation of indexing for half hour timeslots.
- We ensured to round up or round down the half hour time slots if the operating hours do not occur at perfect 30 minute intervals. That is, if operating hours are listed from 9:15 am to 2:05 pm. Then the start time would be rounded up the nearest 30 minute interval and the end time would be rounded down to the nearest 30 minute interval. This means that Sally can reserve rooms in the timeslots between 9:30 am to 2:00 pm.
- We also added some tests to ensure that the helper methods added in the backend work as intended.

## Changes in the Frontend

- Facilitating this change required making changes to the frontend so that we are able to correctly accept the reserved data map passed in from the backend endpoint.
- This involved introducing a helper function to format time based on the operating hour start and number of timeslots provided by the endpoint.

### DIFF
--- a/backend/api/coworking/reservation.py
+++ b/backend/api/coworking/reservation.py
@@ -18,7 +18,7 @@ from ...models.coworking import (
     ReservationMapDetails
 )
 
-__authors__ = ["Kris Jordan"]
+__authors__ = ["Kris Jordan, Yuvraj Jain"]
 __copyright__ = "Copyright 2023"
 __license__ = "MIT"
 
--- a/backend/api/coworking/reservation.py
+++ b/backend/api/coworking/reservation.py
@@ -15,6 +15,7 @@ from ...models.coworking import (
     ReservationRequest,
     ReservationPartial,
     ReservationState,
+    ReservationMapDetails
 )
 
 __authors__ = ["Kris Jordan"]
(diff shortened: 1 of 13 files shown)


yuvrajjain2003:
Resolving the merge conflicts led to something breaking. Please hold on while I try to resolve this! Thanks

KrisJordan (approved):
On first pass this lgtm! Great work moving to dynamic support!

PR #313, opened 2024-03-03
Create Walk-in Reservations for XL Members from Ambassador Welcome Desk

KrisJordan:
Ambassadors requested the ability to create a drop-in reservation for a community member showing up to the XL without having already created a reservation.

This PR adds a new area to the Ambassador Home Component that allows searching for a community member and registering an available seat for them. Since this is staff-only, the UX around pop-up notifications is just alerts for now. Future work could improve this to make use of snackbar or other material UX.

To implement this feature, a few other useful additions were made in other parts of the application, including:

* Search for XL members by partial PID (with unit test coverage)
* Get notified of changes from the `UsersLookup` widget so that components which use it can have callback methods (example use case shown in this widget: after a user is selected, we lookup available seats and show the control for selecting a seat type)


frontend/src/app/coworking/ambassador-home/ambassador-home.component.ts
```diff
@@ -39,15 +61,55 @@ export class AmbassadorPageComponent implements OnInit, OnDestroy {
...
+        .makeDropinReservation(
+          seatSelection,
+          this.welcomeDeskReservationSelection
+        )
+        .subscribe({
+          next: (reservation) => {
+            this.welcomeDeskReservationSelection = [];
+            // Hack to force the pull of new reservations
+            this.ngOnDestroy();
+            this.ngOnInit();
```
ajaygandecha, line 99: Curious, does using `.pollStatus()` here fail?

frontend/src/app/coworking/ambassador-home/ambassador-home.component.css
```diff
@@ -1,7 +1,20 @@
...
+  margin-bottom: 16px;
+}
+
+.walkinReservation.mat-mdc-card-content:last-child {
+  padding-bottom: 0;
+}
+
+coworking-dropin-availability-card:last-child {
+  display: block;
+  margin-bottom: 16px;
```
ajaygandecha, line 15: Nit: This margin did not seem to be active on my computer!

ajaygandecha (approved):
LGTM @KrisJordan!! Added one question and one nit, but this is a great feature that will definitely help out the ambassadors. Like the user lookup too!

frontend/src/app/coworking/ambassador-home/ambassador-home.component.ts
KrisJordan: So pollStatus is pulling in the coworking status (seats available), but we're trying to pull in the current reservations immediately (otherwise it's on a 5 second poll). Since there's already a subscription being tracked in ngOnInit/ngOnDestroy, reusing these makes sense. However, this just gave me an idea for cleaning this up a bit. Going ot make a minor refactor and push again.

frontend/src/app/coworking/ambassador-home/ambassador-home.component.css
KrisJordan: Fair enough, removing for now

PR #322, opened 2024-03-06
Bug Fix and More Padding!

tmattvu:
- Fixed edge case where a newly registered user would have to re-accept community agreement if they click save profile button after accepting community agreement the first time
- Added more padding to the bottom of accept button in dialog box to ensure that the whole button is visible for all layouts of mobile


KrisJordan (approved):
LGTM

PR #323, opened 2024-03-07
299 Add Test Data for Office Hours Tables

meghansun322:
# Summary
This PR adds dummy test data for the office hours table that we've established. More specifically, it adds data for:
- Section Member (Student, Instructor, UTA)
- Office Hours Section
- Office Hours Event
- Office Hours Tickets (Queued, Called, Cancelled, and Closed)
- Association Table Between Ticket and User who created ticket

### Small Model Changes To Have Test Data To Work!
1. We removed `caller_id` from our Ticket Model and removed it from our from_model method
2. Moved `oh_event: OfficeHoursEvent` to the Ticket level rather than the details level
    - to_model() in Office Hours Ticket Didn't Reference `oh_event` At this level, we had the type to be `OfficeHoursEventPartial`, which may not be the behavior we want. We only override this type to `OfficeHoursEvent` in the details class. 
4. Changed field `date` to `event_date` in Office Hours Event 
   - Getting an error due to the name being the same as the type


backend/entities/office_hours/event_entity.py
```diff
@@ -77,7 +77,7 @@ def from_model(cls, model: OfficeHoursEvent) -> Self:
             type=model.type,
             description=model.description,
             location_description=model.location_description,
-            date=model.date,
+            date=model.event_date,
```
meghansun322, line 80: I was getting a type error when running the reset demo script, as the variable name and type name were the same. Changed the name to `event_date` to avoid future problem with naming

backend/models/office_hours/ticket.py
```diff
@@ -39,6 +39,7 @@ class OfficeHoursTicket(OfficeHoursTicketDraft):
     caller_notes: str = ""
     called_at: datetime | None = None
     closed_at: datetime | None = None
+    oh_event: OfficeHoursEvent
```
meghansun322, line 42: Moved this from Details model so that the `oh_event` field in the to_model() method would avoid having to create a `OfficeHoursEventPartial`. I'm not sure if this is the best way to do it as we typically want the relationship fields in the details model. Please let me know thoughts!

sadieamato (commented):
This is looking really good, thanks Meghan! I left a few nit changes to make, as well as a thought on altering the test ticket description if we want to use this data for frontend display purposes. Also, I am also not sure about whether the event field is in the right spot now.

backend/test/services/office_hours/office_hours_data.py
```diff
@@ -0,0 +1,151 @@
...
+    end_time=datetime.now(),
+)
+
+comp_110_oh_events = [comp_110_oh_event_1, comp_110_oh_event_2]
+
+# Ticket For An Event
+pending_ticket = OfficeHoursTicket(
+    id=1,
+    oh_event=OfficeHoursEventPartial(id=1),
+    description="I need help",
```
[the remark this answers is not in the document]
meghansun322: Great point! Will update the description to include the new lines with the different fields!

PR #324, opened 2024-03-08
Room reservation soft launch improvements

KrisJordan:


backend/services/coworking/reservation.py
yuvrajjain2003, line 1: I can fix this later. I know what the issue here is. I was trying to keep SN156 reservations in the date map to gray out slots if someone has a drop-in reservation, but it leads into some problem with the way reservations are drafted. SN156 reservations are drafted with room set as None, so the query doesn't work.

yuvrajjain2003 (approved):
LGTM

PR #328, opened 2024-03-09
Get Organization / Event Test Coverage to 100%

ajaygandecha:
This small PR adds more Pytests to get the tests for `OrganizationService` and `EventService` to 100%.

This PR also removes the unused `OrganizationNotFoundException`, which was causing the exceptions to not hit 100% test coverage.

### DIFF
--- a/backend/test/services/organization/organization_test.py
+++ b/backend/test/services/organization/organization_test.py
@@ -25,6 +25,7 @@ from .organization_test_data import (
     to_add,
     cads,
     new_cads,
+    to_add_conflicting_id,
 )
 from ..user_data import root, user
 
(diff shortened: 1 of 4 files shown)


backend/test/services/organization/organization_test.py
```diff
@@ -81,6 +82,17 @@ def test_create_organization_as_root(organization_svc_integration: OrganizationS
     assert created_organization.id is not None
 
 
+def test_create_organization_id_already_exists(
+    organization_svc_integration: OrganizationService,
+):
+    """Test that the root user is able to create new organizations when an extraneous ID is provided."""
+    created_organization = organization_svc_integration.create(
+        root, to_add_conflicting_id
+    )
```
ajaygandecha, line 91: @KrisJordan when creating this test, I wanted to ask if we want to start standardizing all features to use a `Draft` model to specifically avoid issues with IDs when creating items, including for organizations, events, etc. Thoughts? For now, I added these tests to get to 100% coverage before such a refactor is done more widely.

KrisJordan: I think moving in this direction makes sense, but agree it's a larger refactoring that is orthogonal

KrisJordan (approved):
LGTM

KrisJordan (approved):
LGTM

PR #334, opened 2024-03-09
Fix Academics Home and Section Dropdown to Use Outlined Style

ajaygandecha:
This small PR updates the academics home page card and the section dropdown in the section page to use the Material outlined style.


KrisJordan (approved):
LGTM

PR #337, opened 2024-03-09
Fix Design of Coworking Hours Button

ajaygandecha:
This small pull request properly right-aligns the "All Hours" button on the coworking home page. In addition, since this information would be valuable off-hours, the button now also appears when the XL is closed.


KrisJordan (approved):
LGTM!

PR #339, opened 2024-03-09
Fix Button Casing Throughout CSXL Site

ajaygandecha:
This small PR fixes casing on words in buttons throughout the CSXL site to be more consistent. Most of the buttons used proper case rather than uppercase, so this commit changes some of the final remaining uppercased buttons to proper case.


KrisJordan (approved):
LGTM

Issue #341, opened 2024-03-09
Complete Test Coverage for Room Reservations


yuvrajjain2003:
Currently, the tests don't cover all the functions written that are used for the room-reservation feature. We need to ensure that there is a test for all the different cases for the functions. 

## Acceptance Criteria

- [x] Write tests for all the helper functions for `get_map_reservations_by_date()`
- [x] Write tests for the function used to calculate number of remaining reservation hours. 
- [x] Ensure that there is 100% test coverage and all the tests are passing.

PR #345, opened 2024-03-18
Refactor logic for graying out timeslots

### LINKED ISSUE #340: Refactor Graying out timeslots logic
## Why?

If Sally student has a drop-in reservation, then she should not be able to reserve a room during that time slot. Therefore, the table should gray out these slots to indicate this. Also, if the current time is 11:03 am, then Sally should still be able to reserve a room from 11:03 to 11:30 am. 

## Acceptance Criteria

- [ ] Round down to the nearest half hour instead of rounding up to the nearest half hour when graying out time slots based on current time.
- [ ] Time slots should be grayed out during the time when a person has a drop-in reservation to the XL.
- [ ] Figure out what is wrong with operating hours and why room reservations are opening at 10:30 am rather than 10:00 am.

yuvrajjain2003:
This PR closes #340 and makes some progress towards #341 

-  Round down to the nearest half hour instead of rounding up to the nearest half hour when deleting past timeslots.
- Time slots should be grayed out during the time when a person has a drop-in reservation to the XL.
- Addressed the issue with room reservations table starting at 10:30 am rather than 10:00 am. This was a small edge case that I had missed while implementing the `round_to_closest_half_hour()` helper function, i.e. when minute is exactly 0. 
- We are no longer graying out past timeslots anymore and instead just choose to not display them anymore. This was easy to change since the table is now dynamic and no longer static.

### DIFF
--- a/backend/services/coworking/reservation.py
+++ b/backend/services/coworking/reservation.py
@@ -382,7 +382,9 @@ class ReservationService:
         minutes = dt.minute
 
         if round_up:
-            if minutes < 30:
+            if minutes == 0:
+                to_add = timedelta(minutes=0)
+            elif minutes < 30:
                 to_add = timedelta(minutes=(30 - minutes))
             else:
                 to_add = timedelta(minutes=(60 - minutes))
--- a/backend/services/coworking/reservation.py
+++ b/backend/services/coworking/reservation.py
@@ -323,10 +323,12 @@ class ReservationService:
         for room in rooms:
             time_slots_for_room = [0] * operating_hours_duration
 
-            # Making slots up till current time gray
-            if date.date() == current_time.date():
-                for i in range(0, current_time_idx):
-                    time_slots_for_room[i] = RoomState.UNAVAILABLE.value
+            # # Making slots up till current time gray
+            # This code no longer required, but may be required in the future. 
+            # Please keep this here for now.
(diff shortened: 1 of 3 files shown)


backend/services/coworking/reservation.py
```diff
@@ -519,10 +550,9 @@ def _get_reservable_rooms(self) -> Sequence[RoomDetails]:
         Returns:
             Sequence[RoomDetails]: A sequence of RoomDetails models representing all the reservable rooms, excluding room 'SN156'.
         """
-
         rooms = (
             self._session.query(RoomEntity)
-            .where(RoomEntity.reservable == True)
+            .where(or_(RoomEntity.reservable == True, RoomEntity.id == 'SN156'))
```
KrisJordan, line 555: Let's think of ways of achieving the graying out for drop-in/desk reservations without this logic in the future.

KrisJordan (approved):
LGTM with some future iteration around alignment with drop-in and SN156

backend/services/coworking/reservation.py
yuvrajjain2003: Agreed, this was something that felt uncomfortable to me as well! I have a couple of ideas already in mind that might work with graying out drop-in reservations and circumvent this logic that I will try out this week.

PR #352, opened 2024-03-23
Add special hour blocks for furniture install

KrisJordan:
We need to block off hours to avoid reservations while furniture is getting installed this coming week; this will be reverted and more finite hours will be added later.


ajaygandecha (approved):
LGTM! Ready to merge

PR #373, opened 2024-03-28
357 avoid duplication of polling data in coworking home

### LINKED ISSUE #357: Avoid duplication of polling data in coworking home
The coworking status API reports upcoming reservations. It should not be necessary to also poll for room reservations because that information can be served via the status API endpoint. If there is something missing from the status API endpoint, such as the bound of future reservations is too short (worth checking) then we can fix that. Otherwise, we should pipe off of the status API/service for room reservation data and just filter what comes in. This will cut in half the number of requests going to the backend from the coworking home.

JohnSchachte:
closes #357 

All polling will now take place with the RxObject 'RxCoworkingStatus' to avoid duplicate pulling. 
I also went outside the scope of this branch to refactor how the Room Reservation Service was being utilized and removed duplicate Reservation State Change Methods (like cancel Reservation etc) from this child class in effort to utilize the Reservation Service parent class.

Note:
THe original issue: https://github.com/unc-csxl/csxl.unc.edu/issues/357 mentions 'if there is something missing from the status API endpoint, such as the bound of future reservations is too short (worth checking) then we can fix that.'. I was not sure what this was referring to or how to test if this was an issue. Let me know if this is still worth looking into!

### DIFF
--- a/frontend/src/app/coworking/coworking-home/coworking-home.component.html
+++ b/frontend/src/app/coworking/coworking-home/coworking-home.component.html
@@ -56,7 +56,7 @@
           <h2 class="heading">Upcoming Reservations</h2>
         </div>
         <div class="upcoming-reservations-container">
-          <ng-container *ngFor="let r of initUpcomingReservations() | async">
+          <ng-container *ngFor="let r of upcomingReservations$ | async">
             <ng-container *ngIf="r?.room?.id">
               <coworking-reservation-card
                 [reservation]="r"
--- a/frontend/src/app/coworking/coworking-home/coworking-home.component.html
+++ b/frontend/src/app/coworking/coworking-home/coworking-home.component.html
@@ -51,7 +51,7 @@
     <div class="upcoming-card-container">
       <!-- Show upcoming room reservations -->
       <ng-container
-        *ngIf="((initUpcomingReservations() | async) ?? []).length > 0">
+        *ngIf="((upcomingReservations$ | async) ?? []).length > 0">
         <div class="reservations-header">
           <h2 class="heading">Upcoming Reservations</h2>
         </div>
--- a/frontend/src/app/coworking/coworking-home/coworking-home.component.html
+++ b/frontend/src/app/coworking/coworking-home/coworking-home.component.html
@@ -51,12 +51,12 @@
     <div class="upcoming-card-container">
       <!-- Show upcoming room reservations -->
       <ng-container
-        *ngIf="((upcomingRoomReservation$ | async) ?? []).length > 0">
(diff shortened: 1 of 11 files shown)


JohnSchachte:
Upated the upcoming reservations to be rendering asynchronously with a property in Coworking Home Component class instead of a method call per Kris. Pulled Main and merged with this branch so there should be no conflicts at this time.

frontend/src/app/coworking/coworking-home/coworking-home.component.html
```diff
@@ -51,12 +51,12 @@ <h3 matListItemMeta>&rsaquo;</h3>
...
       <!-- Show upcoming room reservations -->
       <ng-container
-        *ngIf="((upcomingRoomReservation$ | async) ?? []).length > 0">
+        *ngIf="((upcomingReservations$ | async) ?? []).length > 0">
         <div class="reservations-header">
           <h2 class="heading">Upcoming Reservations</h2>
         </div>
         <div class="upcoming-reservations-container">
-          <ng-container *ngFor="let r of upcomingRoomReservation$ | async">
+          <ng-container *ngFor="let r of initUpcomingReservations() | async">
```
KrisJordan, line 59: We should see if we can drop this method call, as well. This is looking really close!

KrisJordan (changes requested):
Almost there; one more refactoring away from piping async over a method call and I think this is done. Glad to see us down to a single network request, this is just an optimization for client-side efficiency purposes.

frontend/src/app/coworking/reservation/reservation.service.ts
```diff
@@ -58,6 +58,18 @@ export class ReservationService {
     );
   }
 
+  checkin(reservation: Reservation) {
+    let endpoint = `/api/coworking/reservation/${reservation.id}`;
+    let payload = { id: reservation.id, state: 'CHECKED_IN' };
```
ajaygandecha, line 63: Nit: Would be good to use an `enum` rather than the string literal `'CHECKED_IN'`. This might be out of scope for this branch though

ajaygandecha (commented):
LGTM, I am going to re-request Kris for a review here.

KrisJordan (commented):
Looking good! I'm having trouble successfully testing on my iPad, so I may need @ajaygandecha to confirm coworking is still functional (drop in and room reservation) from the front-end. Then we can get this merged.

I'm taking a stab at cleaning up one section of code where there was  use  of double not operator to coalesce undefined/null to boolean.

ajaygandecha (approved):
@KrisJordan looks like it is working on my end!

KrisJordan (commented):
LGTM

KrisJordan (approved):

PR #375, opened 2024-03-28
374 active reservations Ambassador View

### LINKED ISSUE #353: Bug fix with active reservations not showing as red on table
## Why?

After making the switch. to dynamic tables, we made the decision to not gray out past timeslots, but instead remove them outright. However, this causes an issue with the logic that was placed in the backend to calculate what timeslots should be shown as red. 

Imagine that it is 2 pm and you make a reservation from 2 - 4pm in SN146. Initially the 4 timeslots will appear red as expected. However, after 30 minutes, these timeslots appear green again, which is a problem since other people can try to reserve these slots and run into unnamed errors. 

This is happening because when populating the red timeslots in the table, I calculate the start_idx and end_idx for the reservation, and with the switch to dynamic tables, the start_idx for the above example becomes -1, which causes some issues with the logic in place.

## Acceptance Criteria

- [x] Resolve the bug and ensure that reserved rooms appear as red even after reservation has been made.

aarjavjain2002:
# !! Active Reservations for AMBASSADOR VIEW (NOT Coworking)!!
This PR updates the filtering criteria on the backend for getting all reservations for the **AMBASSADOR VIEW** (apologies for the branch name, there was a small misunderstanding). Active reservations used to disappear from the Active section after a certain duration. Now all active reservations show up for their full duration and only disappear until the end time is reached or the student checks out/ is checked out.

Yuvraj has also made the following changes:
Fixed the table to show booked out slots when the start time slot has been dynamically purged from the table. This prevents conflicting reservations.

There was a bug where past reservations would have a start time outside the table due to the table being dynamic. This caused the reservation to not show up at all and let people double book the room for the same timeslot. 

This PR will address that bug and closes #353

### DIFF
--- a/backend/services/coworking/reservation.py
+++ b/backend/services/coworking/reservation.py
@@ -463,6 +463,8 @@ class ReservationService:
         """
         office_hours = self._policy_svc.office_hours(date=date)
         for room_id, hours in office_hours.items():
+            if room_id not in reserved_date_map:
+                continue
             for start, end in hours:
                 start_idx = max(self._idx_calculation(start, operating_hours_start), 0)
                 end_idx = min(
--- a/backend/services/coworking/reservation.py
+++ b/backend/services/coworking/reservation.py
@@ -1037,7 +1037,7 @@ class ReservationService:
             self._session.query(ReservationEntity)
             .join(ReservationEntity.users)
             .filter(
-                ReservationEntity.start >= now - timedelta(minutes=10),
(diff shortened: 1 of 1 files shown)


KrisJordan (approved):
Lgtm

KrisJordan:
Actually, although the fixes look good and I'm going to deploy for improving UX, we do need to circle back here this week and add tests that cover these cases to be sure we don't accidentally regress down the road.

PR #400, opened 2024-04-15
399 add checkin time to avoid cancellation verbiage

### LINKED ISSUE #399: Add checkin time to avoid cancellation verbiage
## Why:
Sally the Student might need explanations on what to do at certain states of a reservation. For example, when the reservation is able to check-in and the reservation is in "CONFIRMED", there should be text explaining that the reservation will be cancelled within 10 minutes.

## Acceptance Criteria:
- [ ] User is direct on what times to checkin for their reservation to avoid cancellation

JohnSchachte:
Closes #399 

This is a short PR that will add back the checkin times for confirmed reservations. 
There was also an addition to the "checkinDeadline" method within coworking-card.ts where it makes the checkin deadline deadline the minimum of the start time + 10 and the end of the reservation so the deadline wouldn't display past the reservation's end time.

### DIFF
--- a/frontend/src/app/coworking/widgets/coworking-reservation-card/coworking-reservation-card.html
+++ b/frontend/src/app/coworking/widgets/coworking-reservation-card/coworking-reservation-card.html
@@ -50,6 +50,10 @@
         >
       </h3>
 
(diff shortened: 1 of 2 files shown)


jadekeegan (approved):
LGTM!

KrisJordan (approved):
LGTM

PR #402, opened 2024-04-16
Fall 2024 UTA Application (and admin refactors)

bwgoulet:
1. This PR is for integrating the new TA Application feature into the XL.
2. This PR is for integrating refactors involving event/organization management into the XL


backend/entities/application_entity.py
```diff
@@ -0,0 +1,386 @@
...
+        self.expected_graduation = model.expected_graduation
+        self.program_pursued = model.program_pursued
+        self.other_programs = model.other_programs
+        self.gpa = model.gpa
+        self.comp_gpa = model.comp_gpa
+        self.comp_227 = model.comp_227
+        self.preferred_sections = sections
+
+
+class New_UTA_Entity(UTAEntity):
```
ajaygandecha, line 217: Recommend renaming this to `NewUTAApplicationEntity` - should remove `_` snake case syntax to better match Python conventions!

backend/entities/application_entity.py
```diff
@@ -0,0 +1,386 @@
...
+            **parent_model,
+            intro_video=self.intro_video,
+            prior_experience=self.prior_experience,
+            service_experience=self.service_experience,
+            additional_experience=self.additional_experience,
+        )
+
+    def update(self, model: New_UTADetails, sections: list[Section]) -> None:
+        """
+        Update an ApplciationEntity from a New_UTA model.
```
ajaygandecha, line 294: Typo: `ApplciationEntity` -> `ApplicationEntity`

backend/entities/academics/section_entity.py
```diff
@@ -79,6 +74,12 @@ class SectionEntity(EntityBase):
         primaryjoin="and_(SectionEntity.id==SectionMemberEntity.section_id, SectionMemberEntity.member_role!='STUDENT')",
     )
 
+    # All applicants where section is preferred
+    # NOTE: This field establishes a many-to-many relationship between the sections and applications table.
+    preferred_applicants: Mapped[list["UTAEntity"]] = relationship(
+        secondary=section_application_table, back_populates="preferred_sections"
```
ajaygandecha, line 80: This relationship might lead to an issue over time and as more terms pass, since this list would not contain only applicants in the most recent application cycles. I recommend taking a look at the `primaryjoin=` examples used in the relationship fields defined above to filter this to only include future terms.

backend/models/application_details.py
```diff
@@ -0,0 +1,64 @@
...
+class Returning_UTADetails(Returning_UTA):
+    """
+    Pydantic model to represent a `UTA Application`.
+
+    This model is based on the `UTA` model, which defines the shape
+    of the `Application` database in the PostgreSQL database.
+    """
+
+
+class UserApplication(BaseModel):
```
ajaygandecha, line 55: Wondering if this model is redundant if it is just a wrapper for `New_UTADetails`! Also, should this return new details always, even when students are returning applicants?

backend/services/application.py
```diff
@@ -0,0 +1,230 @@
...
+        #             entity.to_details_model()
+        #         )
+        #     elif entity.type == "returning_uta":
+        #         applications.append(entity.to_details_model())
+        #     elif entity.type == "uta":
+        #         applications.append(entity.to_details_model())
+        #     else:
+        #         applications.append(entity.to_model())
+
+        # return applications
```
ajaygandecha, line 65: Leftover code to delete

backend/services/application.py
```diff
@@ -0,0 +1,230 @@
...
+        
+    
+    def get_application(self, subject: User) -> UserApplication:
+        """Returns application(s) for a specific user
+
+        Returns:
+            list[Application]: List of all current applications for a specific user
```
ajaygandecha, line 72: Check this docstring!

backend/entities/section_application_table.py
```diff
@@ -0,0 +1,19 @@
...
+from .entity_base import EntityBase
+
+
+__authors__ = ["Abdulaziz Al-Shayef, Ben Goulet"]
+__copyright__ = "Copyright 2023"
+__license__ = "MIT"
+
+# Define the section_application table to be used as a join table to persist a
+# many-to-many relationship between the section and application table.
+section_application_table = Table(
```
ajaygandecha, line 13: I think are moving towards a new convention where association tables are modeled as entities rather than as `Table` objects, so it might be good to convert this to an entity - will ping @KrisJordan to confirm this

backend/services/application.py
```diff
@@ -0,0 +1,230 @@
...
+
+        section_ids = [section[1] for section in sections]
+
+        sections_entity = []
+        for id in section_ids:
+            sections_entity.append(
+                self._session.query(SectionEntity)
+                .filter(SectionEntity.id == id)
+                .first()
+            )
```
ajaygandecha, line 99: Small nit: Is it possible to do a more efficient search here? It should be possible to query based on a list of IDs, then remove duplicates (if needed) - this will help to limit the number of queries. Also, it might be better to break up this expression

backend/services/application.py
```diff
@@ -0,0 +1,230 @@
...
+
+        section_dict = {
+            i: section.to_model() for i, section in enumerate(sections_entity)
+        }
+
+        application = application_entity.map_application_to_detail_model(section_dict)
```
ajaygandecha, line 105: Do we need the `map_application_to_detail_model` function + is the data not automatically queried from the database?

Also, should the return type of this be just `New_UTA_Details`?

backend/services/application.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def __init__(self, session: Session = Depends(db_session)):
+        """Initializes a new ApplicationService.
+
+        Args:
+            session (Session): The database session to use, typically injected by FastAPI.
+        """
+        self._session = session
+
+    def list(self) -> list[New_UTADetails]:
```
ajaygandecha, line 39: Adding this comment here because reading through this file, I was a little concerned with return types! Frontend implementation might remedy this (so no worries!), but all of these services are returning `New_UTADetails` - in the case that an application is for a returning member, is that data lost / hidden once this data is transferred to the frontend?

backend/services/application.py
```diff
@@ -0,0 +1,230 @@
...
+        section_dict = {
+            i: section.to_model() for i, section in enumerate(sections_entity)
+        }
+
+        application = application_entity.map_application_to_detail_model(section_dict)
+
+        return UserApplication(application=application)
+    
+
+    def create_undergrad(self, application: New_UTADetails) -> New_UTADetails:
```
ajaygandecha, line 110: Nit: Would probably rename to `create_uta_application`

backend/services/application.py
```diff
@@ -0,0 +1,230 @@
...
+                .values(preference=index)
+            )
+
+        self._session.commit()
+
+        return application_entity.to_details_model()
+
+    def update_undergrad(
+        self, subject: User, application: New_UTADetails
+    ) -> New_UTADetails:
```
ajaygandecha, line 158: Same thing here, I think this update function could potentially wipe returning user data

backend/services/application.py
```diff
@@ -0,0 +1,230 @@
...
+                .filter(
+                    section_application_table.c.application_id
+                    == original_application.id,
+                    section_application_table.c.section_id == section.id,
+                )
+                .values(preference=index)
+            )
+        self._session.commit()
+
+        return original_application.to_details_model()
```
ajaygandecha, line 212: Nit: Convention is to move away from `.execute()` in favor of direct session method calls, this would also make the calls in this section more readable!

backend/api/application.py
```diff
@@ -0,0 +1,129 @@
...
+
+openapi_tags = {
+    "name": "Applications",
+    "description": "Create, update, delete, and retrieve TA Applications.",
+}
+
+api = APIRouter(prefix="/api/applications")
+
+
+@api.get("", response_model=list[New_UTADetails], tags=["Applications"])
```
ajaygandecha, line 29: See comment in services section about return types

backend/services/application.py
```diff
@@ -0,0 +1,230 @@
...
+        """Returns application(s) for a specific user
+
+        Returns:
+            list[Application]: List of all current applications for a specific user
+        """
+
+        application_entity = (
+            self._session.query(ApplicationEntity)
+            .filter(ApplicationEntity.user_id == subject.id)
+            .first()
```
ajaygandecha, line 78: For this to work, make sure that this is sorted! This might cause issues as users apply in many terms. In addition, there are some edge cases to consider:
* Users may have active applications for say, Summer Session I and Summer Session II at the same time.
* Users might take a break from being a TA and not submit an application for the current term, but they may have applied in the past. Should this function return the past application (i.e., the most recent one), or `None` if they did not apply to an application for currently active terms?

ajaygandecha (changes requested):
Amazing job so far @alshayefaziz and @bwgoulet! This CR just comments on the backend first -- once the backend review completes, I can continue with the frontend. Let me know if you have any questions!

backend/entities/application_entity.py
```diff
@@ -0,0 +1,386 @@
...
+    other_programs: Mapped[str] = mapped_column(String, nullable=True)
+
+    # GPA
+    gpa: Mapped[float] = mapped_column(Float, nullable=True)
+
+    # COMP GPA
+    comp_gpa: Mapped[float] = mapped_column(Float, nullable=True)
+
+    # Do they want to do this as COMP 227?
+    comp_227: Mapped[str] = mapped_column(String, nullable=True)
```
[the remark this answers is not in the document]
alshayefaziz: This would not be a bool because the options vary on the semester. I.e:
1. Not taking COMP 227
2. Taking COMP 227 (no compensation)
3. Doing a split
With the options potentially changing in the future, we decided to store this as a string. Do you think we should change it to something like an ENUM?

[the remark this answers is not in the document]
ajaygandecha: @alshayefaziz This might be better off as an `Enum`! A good example for this is `RosterRole` in the academics feature

frontend/src/app/ta-application/ta-application.service.ts
```diff
@@ -0,0 +1,128 @@
...
+    );
+  }
+
+  deleteApplication(): void {
+    this.http.delete('/api/application/delete');
+  }
+
+  getProfile(): Observable<Profile> {
+    return this.http.get<Profile>('/api/profile');
+  }
```
ajaygandecha, line 115: We can use the profile already injected to avoid duplicated code! See the other services for inspiration.

frontend/src/app/ta-application/ta-application.service.ts
```diff
@@ -0,0 +1,128 @@
...
+  getApplication(): Observable<Application | null> {
+    return this.http
+      .get<Application | null>('/api/applications/user')
+      .pipe(
+        tap((application) => console.log('Fetched application:', application))
+      );
+  }
+
+  submitApplication(
+    application: Omit<Application, 'id'>
```
ajaygandecha, line 72: Aside: What is `Omit`? We have never used this, seems intriguing

frontend/src/app/shared/slack-invite-box/slack-invite-box.widget.ts
```diff
@@ -0,0 +1,22 @@
+import { Component, OnInit } from '@angular/core';
```
ajaygandecha, line 1: This file might cause a merge conflict - I think this is already merged in.

frontend/src/app/admin/applications/list/admin-applications-list.component.ts
```diff
@@ -0,0 +1,61 @@
+/**
+ * The Admin Application List page retrieves and displays a list of
+ * all pending TA application and provides functionality to review them.
```
ajaygandecha, line 3: This component seems to not be implemented, so it would be good to avoid merging this in.

frontend/src/app/ta-application/ta-application.service.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  /** Returns a list of all Applications
+   * @returns {Observable<Application[]>}
+   */
+  list(): void {
+    this.http
+      .get<Application[]>('/api/applications')
+      .subscribe((applications) => this.applications.set(applications));
+  }
```
ajaygandecha, line 54: Since this is in the Admin Service, we can probably remove it here

frontend/src/app/admin/applications/widgets/application-review-modal.widget.ts
```diff
@@ -0,0 +1,51 @@
+import { Component, Inject, OnDestroy } from '@angular/core';
+import { MAT_DIALOG_DATA } from '@angular/material/dialog';
+import { HttpClient } from '@angular/common/http';
+import { Application } from '../admin-application.model';
+import { MatSnackBar } from '@angular/material/snack-bar';
```
ajaygandecha, line 5: This also does not seem to be implemented, am I correct? I could not find this in the admin page.

ajaygandecha (changes requested):
@alshayefaziz @bwgoulet, this is my frontend CR since Kris said there is a time crunch!

Also, there is a major organizations feature that seems to be lumped into this PR - I did not review this, I think it would be better for this to be a separate pull request rather than put in this branch. It is too large of a feature to be put in the same branch as this one, and I want to focus on UTA applications for now when reviewing!

Lmk if you have any questions about this review! Looking great so far, love the use of the mat stepper!!

frontend/src/app/shared/slack-invite-box/slack-invite-box.widget.ts
ajaygandecha: Nevermind, it was not! Again, as I mentioned, this PR contains too many new features, so it would probably be good to move this out of the PR for UTA application so it can be reviewed separately later

frontend/src/app/ta-application/ta-application.service.ts
bwgoulet: Sends the Application object without an id field - I use it here to avoid issues with assigning an id for the frontend model, allowing the backend to set the id instead!

backend/models/application_details.py
alshayefaziz: Yes it is redundant! That is my fault, I had it set up like this just in case I had to approach ordinality issues in a different manner and return 'preferred sections' in the model with the application. Will remove though.

jadekeegan (changes requested):
Backend review! Most of my comments are just minor documentation fixes, but I did notice a few more pressing issues:
1. There are no permissions in any of the service methods, which means any user can create an application for another user, view ALL applications, update another user's applications, and delete another user's applications.
2. I'm not sure if there is just no use case for this yet, but it seems that there are also no admin capabilities regarding applications. Once permissions are fixed according to (1), admin users won't be able to get/create/update/delete user applications, so a special check will need to be implemented for this.
3. As Ajay also mentioned, there should probably be more nuance to retrieving applications for all the service methods rather than just getting the first one that pops up for the given user!

EDIT: Seeing that there are no admin capabilities on the frontend, but definitely still update the backend to avoid security concerns!! The fixes should all be fairly straightforward, but they will require some refactoring of the service methods and APIs. I think Event may have some similar permissions in it if you need something to reference.

frontend/src/app/academics/academics-home/academics-home.component.html
```diff
@@ -27,3 +27,31 @@
     </button>
   </mat-card-actions>
 </mat-card>
+
+<div
+  *ngIf="applicationService.user_application$ | async as application; else noApplication">
+  <mat-card class="content" appearance="outlined">
+    <mat-card-content>
+      <mat-card-subtitle>Become a TA!</mat-card-subtitle>
```
jadekeegan, line 35: The subtitle is a little small, so it may be nice to make this a title rather than a subtitle.

frontend/src/app/admin/applications/admin-applications.service.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * The Admin Organization Service abstracts backend calls from the
```
jadekeegan, line 2: Incorrect documentation.

jadekeegan (changes requested):
Aside from the things Ajay mentioned, I mainly just noticed some documentation issues! Definitely go through the files again and ensure everything is accurate to the file.

frontend/src/app/academics/academics-home/academics-home.component.html
```diff
@@ -27,3 +27,31 @@
     </button>
   </mat-card-actions>
 </mat-card>
+
+<div
```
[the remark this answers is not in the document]
bwgoulet: Replying to review this later when term becomes relevant

backend/entities/section_application_table.py
alshayefaziz: Yes! This is part of the refactor that we plan on doing after launch!

PR #405, opened 2024-04-19
TA Application Backend Fix

alshayefaziz:
This branch currently contains fixes recommended by the CR, it will be updated and pushed to until resolving all edits.


backend/api/application.py
```diff
@@ -23,13 +25,13 @@
     "description": "Create, update, delete, and retrieve TA Applications.",
 }
 
-api = APIRouter(prefix="/api/applications")
+api = APIRouter(prefix="/api/ta/applications")
```
ajaygandecha, line 28: Super tiny nit: I wonder if `/api/applications/ta` might be better here, since we could see `/api/applications/ambassador` in the future!

backend/entities/application_entity.py
```diff
@@ -8,15 +8,20 @@
...
 
 from .entity_base import EntityBase
 from typing import Self, Dict
-from ..models.application import Application, UTA, New_UTA, Returning_UTA
-from ..models.application_details import UTADetails
+from ..models.application import (
+    Application,
+    UTAApplication,
+    NewUTAApplication,
+    Returning_UTA,
```
ajaygandecha, line 22: Nit: Might need to change the name of this too (although I know this is not really used yet)!

ajaygandecha (approved):
Looks amazing @alshayefaziz!! Nice work, thanks for tolerating the rather long CR!

Two small nits, but other than that, LGTM and ready to merge back into the TA branch!

PR #406, opened 2024-04-21
Refactor All Tests to use Core Data, Improve Testing Documentation

ajaygandecha:
This pull request modifies all of the existing pytests to use the `core_data.py` file, which provides a single fixture for all test data for the database to load at once. Previously, some files used `core_data`'s fixture, while others manually imported fixtures directly. This manual importing often caused confusion, especially for students in 423! Fully embracing the `core_data` model will improve code readability and make testing easier to set up.

This PR also adds to the testing docs page to provide an explanation of how fixtures work. This new documentation might also be good to send out to students for Sprint 02.


backend/test/services/core_data.py
```diff
@@ -22,6 +25,13 @@ def setup_insert_data_fixture(session: Session):
     permission_data.insert_fake_data(session)
     organization_test_data.insert_fake_data(session)
     event_test_data.insert_fake_data(session)
+    room_data.insert_fake_data(session)
+    operating_hours_data.insert_fake_data(session, time.time_data())
```
ajaygandecha, line 29: The only piece causing a lot of errors here was getting the `time` fixture to be recognized. Despite importing it in this file and numerous other files, there was still an error complaining about "missing parameter `time`". The only way for me to get around this was manually passing in `.time_data()` into these functions directly. Thoughts @KrisJordan?

ajaygandecha:
Also, future consideration: We could refactor further to make a simple `setup.py` which includes the `conftest` and `core_data` fixtures in one, so test files only have to include this one import line.

KrisJordan:
I like the effort to simplify testing files, however this approach has one significant drawback to the current approach which I believe we should address as the site continues to scale with new features: tests are able to load only the data they need. As these tests are more integration-like than unit-like, because we are testing integration with postgres and all data makes its way to postgres, there is a cost to loading data beyond what is needed for a test. It seems to me within a given feature (academics, office hours, coworking) there are known data fixtures which we could boil down to a single file within the given feature and achieve effectively a very similar outcome while not loading data unrelated to a feature unnecessarily for each test. Thoughts?

ajaygandecha:
@KrisJordan I agree! After revisiting this, especially with the office hours feature, I think I will close this PR. I will open a ticket to refactor test data to how you described above, I think that makes more sense!

Issue #158, opened 2023-11-10
Add Markdown Support to Event Details + Improve Event Details Card UX


jadekeegan:
# Improve Event Details Card UX
Fix the Event Details card styling to prevent text from running off the card. Add markdown to the event details card.

## Why?
These changes will ensure that the events and their information are displayed neatly and cleanly to the users. It will also provide organization leaders with improved features for writing their event descriptions!

## Acceptance Criteria
- [x] Fix text overflowing off of Event Details Card.
- [x] Add markdown capability to the description for the Event Details card.

ajaygandecha:
Moved to #514

PR #407, opened 2024-04-22
Add Markdown Processing to Event Description (#158)

jadekeegan:
This PR adds markdown processing to the event descriptions using installed `marked` and `dompurify` modules. `marked` is used to parse the event description and convert it to HTML, while `dompurify` is used to sanitize the produced HTML.

## Developer Concerns
- Should we use `Text` instead of `String` for our event description to allow for more characters in the description? String is restricted to 255 I believe. (Didn't want to make this change yet as (I think) it would require a small migration)
- May need to edit global styles for things to ensure that colors/etc look good still!

### DIFF
--- a/frontend/src/app/event/event-details/event-details.component.ts
+++ b/frontend/src/app/event/event-details/event-details.component.ts
@@ -19,6 +19,8 @@ import {
 import { Event } from '../event.model';
 import { Observable, of } from 'rxjs';
 import { PermissionService } from 'src/app/permission.service';
+import * as marked from 'marked';
+import * as DOMPurify from 'dompurify';
 
 /** Injects the event's name to adjust the title. */
 let titleResolver: ResolveFn<string> = (route: ActivatedRouteSnapshot) => {
@@ -47,6 +49,7 @@ export class EventDetailsComponent {
 
   /** Store Event */
   public event!: Event;
+  public eventDescriptionHTML: string = '';
 
   /** Store the currently-logged-in user's profile.  */
   public profile: Profile;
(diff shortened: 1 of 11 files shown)


frontend/src/app/event/event-details/event-details.component.ts
```diff
@@ -63,6 +66,9 @@ export class EventDetailsComponent {
     };
     this.profile = data.profile;
     this.event = data.event;
+    this.eventDescriptionHTML = DOMPurify.sanitize(
+      marked.parse(this.event.description) as string
+    );
```
ajaygandecha, line 71: I realize, this might be a good first use case where the Angular Directive construction might be useful. I wonder if we should give it a shot. The directive can then be applied to any Angular text box like this one and can then easily expand markdown support.

Here is some [documentation](https://angular.io/guide/attribute-directives) about directives. They are super useful, because we could end up with something like:

```
<div markdown>{{ input }}</div>
```

We might be expanding our usage of directives in the future, but it is a good place to start.

ajaygandecha (changes requested):
LGTM in its current state! I added a comment about Angular Directives - I think it might be worth giving this a shot, as this would make markdown processing universal across all components.

Also, if you make this directive, export it from the shared module!

ajaygandecha:
@jadekeegan hey, since we have diverged significantly in production since this commit, I am going to close this pull request. However, I will keep you on #407 and we can keep this branch on record for solving this problem in the future!

I created a new ticket #514 which includes more details!

PR #446, opened 2024-05-15
Allow root to create OH Sections

ajaygandecha:
This PR allows the root user to create office hour sections so that there is an API endpoint to add an instructor to the OH section.


ajaygandecha:
Added the API endpoint, all tests pass with 100% coverage.

KrisJordan (approved):
LGTM

PR #454, opened 2024-05-17
[DRAFT] [DO NOT MERGE] Angular v17 Upgrade: Organizations Refactor

ajaygandecha:
This draft pull request catalogs the changes and choices made when refactoring the organizations feature in the CSXL site to new Angular v17 standards. This is the first such refactor, and all of the features will go through the same refactor. The goals of this refactor are to make the existing code significantly more concise, as well as to transition towards using Angular v17 features such as the `Signal<>` state control workflow and other changes.

## File Format Changes and Angular v17 Feature Considerations

The existing directory layout is not optimal. Here is a preview of the current directory layout (in red on the right), and the new proposed directory layout on the left:

img_3300_720

The new structure for ***all*** features will become as follows:
```
feature
|  sample page (and so on...)
|  |  (HTML / CSS / TS)
|  widgets
|  |  sample widget (and so on...)
|  |  |  (HTML / CSS / TS)
| pipes
|  |  sample pipe (and so on...)
|  feature.service.ts
|  feature.models.ts
|  feature.modules.ts
|  feature-routing.modules.ts
```

This leads to a few considerations, as listed:

* **We will now only have one service per feature**. In the existing organizations service, there are three services, some of which have redundant functionality. We will aim to collapse this down to one service where applicable. In situations where a feature contains sub-features of large size (namely only `coworking` and `academics`), they can be separated based on their respective features. Mainly, we will no longer separate admin and permission services from the main service.

* **We can preserve state using signals.** With the introduction of `Signal<>` in Angular v17, and given that resolvers load data prior to the page loading (causing delays), we will slowly be moving away from using resolvers for retrieving reactive data:

  **OLD (With resolvers)**:
  ```ts
  /** Initialize data from resolvers. */
  const data = this.route.snapshot.data as {
    profile: Profile;
    organizations: Organization[];
  };
  
  this.profile = data.profile;
  this.organizations = data.organizations;
  ```
  
  **NEW (Proposed, using Service-emitted Signals)**:
  ```ts
  /** Initialize data from signals **/
  this.profile = this.profileService.profile();
  this.organizations = this.organizationService.organizations();
  ```

* **Pipes are now in their own folder**. Instead of having their own folder with a feature name, pipes are moved into a common folder (like widgets).

* **Admin features will be moved out to an Admin module.** As @KrisJordan mentioned, it might be better for all admin features to be moved to a separate admin folder rather than individual feature folders. Since services are not declared in modules and are injected from root, feature services would be available to the Admin module. This might allow for our file organization to be cleaner.

## Adopting Signals in the Components 

Currently, the service for the organizations feature returns RxJS `Observable` objects, the result of HTTP method calls, for which the components subscribe to. This, along with the `RxObject` construction (`ReplaySubject` + `Observable`), handle state and data transfer in the CSXL application.

The latest version of Angular has announced the `Signal<>` construction which allows us to more easily handle state in our application. There is also RxJS-Signal interoperability (in v17+) which allows us to easily convert to and from Observables to Signals.

Signals have a much more concise syntax and are significantly easier to use. Signals combine the functionality of traditional observables with the reactive functionality we manually created in the `RxObject` abstraction.

**Sample Component TypeScript File**:
```ts
// OLD: With observables
organizations$: Observable<Organization[]>;
organizations: Organization[];
...
this.organizations$ = this.organizationService.getOrganizations();
this.organizations$.subscribe((organizations) => {
  this.organizations = organizations;
};

// NEW: With signals
organizationsSignal: Signal<Organization[]>;
organizations: Organization[];
...
this.organizationsSignal = this.organizationService.getOrganizations();
this.organizations = this.organizationsSignal();
```

In addition, working with signals in the HTML files much easier as well:

**Sample Component HTML File**:
```html
<!-- OLD: Observables -->
<p>{{ (event$ | async).name }}</p>

<!-- NEW: Signals -->
<p>{{ event().name }}</p>
```

Signals also are reactive. This is extremely useful, because in our *service*, we can recreate the usage of `RxObject` very easily:

```ts
@Injectable({ providedIn: 'root' })
export class NewOrganizationService {

  private organizationsSignal: WritableSignal<Organization[]> = signal([]);
  organizations = this.organizationsSignal.asReadonly();
  ...
  createOrganization(organization: Organization): Signal<Organization | undefined> {
    return toSignal(
      this.http.post<Organization>('/api/organizations', organization)
        .pipe(tap((organization) => this.organizationsSignal.update((organizations) => [...organizations, organization])))
      );
  }
```

As you can see, the `createOrganization` function ***returns a signal*** containing the newly created organization, and ***updates an internal organizations list*** to include the newly-created organization! The `organizations` signal acts just as `rxOrganizationsList` would have worked in the old system.

## Adopting Angular v17 in the HTML Files

One interesting update and convention change in the Angular v17 version is how the built-in `ngFor` and `ngIf` directives work. They have added new, built-in control flow into the HTML, which is the new standard. The new standard helps to limit the number of operations needed to modify the DOM (so more efficient), and may be more readable. Here is an example of the old if-else statement vs the new one in the HTML file:

```html
<!-- OLD (Pre Angular v17) -->

<p *ngIf="authenticated; else showAlert">Logged in!</p>
<ng-template #showAlert>
  <p>Not logged in!</p>
</ng-template

<!-- NEW (Angular v17+) (yes, this is HTML code!) -->
@if (authenticated) {
  <p>Logged in!</p>
} @else {
  <p>Not logged in!</p>
}
```

This supports `@if`, `@else if`, and `@else` if-else-if-else statements, `@for` for for-loops, and `@switch` for switch-case statements. This is likely the convention we will be moving towards in Angular v17.


ajaygandecha:
Closing this draft pull request to create the real one.

PR #456, opened 2024-05-18
Create Frontend Pagination Abstraction

ajaygandecha:
This pull request creates a new frontend abstraction for pagination. The abstraction supports simple pagination, as well as pagination where the HTTP response model does not match the type of model stored in each page. The abstraction makes strong use of generic types so that the feature is flexible to be used across the site with minimal localization.

This pull request does not implement the abstraction in any features. I was planning on utilizing this in the events feature refactor.

*Again, I am adding @jadekeegan and @atoneyd to this PR, since we have all been working on pagination for quite a while. No need to complete CRs here! You all should not be bothered with notifications after this PR and the events refactor PR come in!*

## Usage

This PR adds the `Paginator<T, Params>` class, which is the main "service" tool responsible for pagination. This paginator object has two generic types, where `T` is for the type of object being paginated, and `Params` specifies the specific type of parameters used for pagination. This is useful for certain features such as the events feature, where `EventPaginationParams` has a different shape than `PaginationParams` to account for date ranges. The object's constructor takes in an API endpoint as a string, which is the API endpoint to be called every time a new page loads. 

This is an example usage for a hypothetical organization paginator:

```ts
let paginator: Paginator<Organization, PaginationParams> = new Paginator<>('api/organizations');
```

Pages are loaded with the `.loadPage(params)` function, which takes in a params object corresponding with the generic type `Params` provided in the paginator's construction. This is an example usage of loading the next page for the paginator defined above:

```ts
paginator.loadPage(params);
```

When a page is loaded, a *signal* (Angular v17 feature, analogous to `RxObject`) called `page` is updated inside the paginator object. We can easily access the data like so:

```ts
paginator.page(); // Loads the current page.
```

The `loadPage()` function is more powerful. There are many cases in the CSXL site where the *response model* from the API is **not** the same as the model of data we are trying to store in pages. This is most notable in the events feature. The API endpoint for events return data in the `EventJson` format, where dates are codified as strings. Then, in the frontend, this is converted to standard `Event` format. This convention is used so that dates are converted properly.

The `loadPage()` function can account for this. The function has an optional generic type parameter called `APIType`, which by default is just `T`. This generic type signals the type of the response rather than the type of data stored in the page. This can be different from `T` in cases where they are not the same.

In addition, this defines a type alias called `PaginationOperatorFunction<APIType, T, Params>`, which is just an operator function that helps to convert `Paginated<APIType, Params>` objects to `Paginated<T, Params`>. This alias just helps the code to be a bit more readable, and is defined as follows:

```ts
type PaginationOperatorFunction<APIType, T, Params> = OperatorFunction<Paginated<APIType, Params>, Paginated<T, Params>>;
```

So, `.loadPage<APIType>(params, operator) can take in an operator function of type `PaginationOperatorFunction` which processes the conversion.

Here is the example usage for events:
```ts
let paginator: Paginator<Event, EventPaginationParams> = new Paginator<>('/api/event');
paginator.loadPage<EventJson>(params, eventJsonToEventOperatorFunction);
paginator.page(); // Returns the loaded page, in type `Paginated<T>`
```

### DIFF
--- a/frontend/src/app/pagination.ts
+++ b/frontend/src/app/pagination.ts
@@ -99,6 +99,7 @@ abstract class PaginatorAbstraction<T, Params extends URLSearchParams> {
    * @template APIType: (Optional) Response model from the API call, if it is different than `T`.
    * @param paramStrings: Pagination parameters.
    * @param operator: (Optional) Function to convert data from `Paginated<APIType>` to `Paginated<T>`.
+   * @returns {Observable<Paginated<T, Params>>}
    */
   loadPage<APIType = T>(
     paramStrings: Params,
--- a/frontend/src/app/pagination.ts
+++ b/frontend/src/app/pagination.ts
@@ -52,9 +52,6 @@ export interface Paginated<T, ParamType> {
  *
  * @template T: Type of data stored in the paginator's pages.
  * @template Params: Type of the pagination params used by the type of object being paginated.
- *
- * Sample Type:
- * `Paginator<Event, EventPaginationParams>`: Paginates `Event` models using params defined in `EventPaginationParams`.
  */
 abstract class PaginatorAbstraction<T, Params extends URLSearchParams> {
   /** Stores the previously used parameters for reference. */
@@ -84,7 +81,6 @@ abstract class PaginatorAbstraction<T, Params extends URLSearchParams> {
    *
    * Usage:
    * ```
-   * let paginator: Paginator<Organization, OrganizationPaginationParams> = new Paginator<>('api/organization');
    * paginator.loadPage<>(params);
    * paginator.page(); // Returns the loaded page.
    * ```
@@ -96,7 +92,6 @@ abstract class PaginatorAbstraction<T, Params extends URLSearchParams> {
    *
    * Usage:
    * ```
-   * let paginator: Paginator<Event, EventPaginationParams> = new Paginator<>('/api/event');
    * paginator.loadPage<EventJson>(params, parseEventJson);
    * paginator.page(); // Returns the loaded page, in type `Paginated<T>`
    * ```
--- a/frontend/src/app/pagination.ts
+++ b/frontend/src/app/pagination.ts
@@ -139,8 +139,10 @@ abstract class PaginatorAbstraction<T, Params extends URLSearchParams> {
   }
 }
 
+/** Default paginator implementation. */
 export class Paginator<T> extends PaginatorAbstraction<T, PaginationParams> {}
 
+/** Paginator implementation for working with time ranges. */
 export class TimeRangePaginator<T> extends PaginatorAbstraction<
   T,
   TimeRangePaginationParams
--- a/frontend/src/app/pagination.ts
+++ b/frontend/src/app/pagination.ts
@@ -12,7 +12,7 @@
 
 import { HttpClient } from '@angular/common/http';
 import { WritableSignal, signal } from '@angular/core';
-import { OperatorFunction, tap } from 'rxjs';
+import { map, tap } from 'rxjs';
 
 /** Defines the general model for the pagination parameters expected by the backend. */
 export interface PaginationParams {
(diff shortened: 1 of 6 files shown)


frontend/src/app/pagination.ts
```diff
@@ -13,14 +31,123 @@ export interface EventPaginationParams {
...
+      this.http.get<Paginated<APIType, Params>>(route).pipe(
+        operator,
+        tap((pageData) => this.pageSignal.set(pageData))
+      );
+    } else {
+      // Otherwise, just call the API and update the signal.
+      this.http
+        .get<Paginated<T, Params>>(route)
+        .pipe(tap((pageData) => this.pageSignal.set(pageData)));
+    }
```
KrisJordan, line 151: Do these not need to be returned so that they can be subscribed to?

KrisJordan (commented):
Quick comment, didn't get to fully review yet...

frontend/src/app/pagination.ts
ajaygandecha: I do not believe so, since they are only used to update the internal signal! This follows the same convention we have used in admin services with `RxObject`, where the `.get()` methods just update the internal state rather than returning an observable.

ajaygandecha: @KrisJordan now, if we instead wanted the services themselves to hold the signal for the data, then we could just return the observable and use `tap` in the service to update its signal - that would work too!

ajaygandecha:
@KrisJordan I just made a refactor. I removed the `PaginationOperatorFunction` as described above, because this converts from `Paginated<APIType, Params>` to `Paginated<T, Params>`. This means that a new function would need to be created for each type of pagination, which did not seem to be the best solution.

Instead, now `.loadJson()` just takes in an optional function of type `(_: APIType) => T`, which is the same type signature as something like `parseEventJson`. This means that the usage is truly as described, where we could just call:

```ts
let paginator: Paginator<Event, EventPaginationParams> = new Paginator<>('/api/event');
paginator.loadPage<EventJson>(params, parseEventJson);
paginator.page(); // Returns the loaded page, in type `Paginated<T>`
```

KrisJordan (commented):
I applaud the efforts to work toward a better pagination strategy!

This PR feels like it needs to go a step or two further to prove that the refactored `pagination.ts` is, indeed, a win. As it stands, the PR's impacted files outside of `pagination.ts` all see a slight increase in complexity (an additional type parameter added) and it's unclear where the gain comes from (though, the PR's text shows example uses of the higher-level API).

One thing that sets off a minor smell test for me is whether this extra abstraction (Pagination Param Types) should actually be advertised and visible _to_ the code depending on `pagination.ts` or whether it should be hidden as implementation beauty (clean and avoids redundant code). The  counter-proposal here would be to have concrete subclasses per paginationparams type, which extend from the generic class and substitute in the param type. My feeling is there will be a very small number of param types (beyond the offset/limit-based and date range-based we have, I'm not coming up with anything obvious).

While it may seem like the difference between `Paginator<Event, EventPaginationParams>` and `DateRangePaginator<Event>` feels insignificant, it has less conceptual overhead (the extra type parameter leaves the user asking "should I be using something else here? What else _could_ I use here?") and has less code overhead (you should be able to avoid importing `EventPaginationParams` in consumer code).

So those are my two suggestions:

1. Go ahead and prove out the new API's glory in actual use cases in the code
2. Consider hiding some of this abstraction as abstraction detail behind a simpler, more concrete end-user API for the specific types of pagination we have

KrisJordan (commented):
One other comment-- is this the preferred way to get a Signal from an `Observable`?

My heart hurts a little not getting an `Observable` out of this.

I'm starting to wonder whether all signal conversion really should be in the component layer and services remain `Observable` forward.

ajaygandecha:
@KrisJordan thank you for the review here! I completely agree - while the original intention was to have full flexibility with generics, you are right that there is really only two use cases, so this is probably overkill (and hampers code readability). I really like the solution you came up with, so I will refactor to get to that point!

Also, I agree about this abstraction returning observables instead of having a signal - it is probably best that we standardize to services containing the signal (the replacement for `RxObject`), as we have done in services previously. In the service, we can subscribe or tap into the value returned from this abstraction's `.loadPage()` function, and that will work better (and probably be more readable).

Will refactor and then request review again!

ajaygandecha:
Alright @KrisJordan, I performed an initial refactor! What bothered me a little was the duplicated code between `Paginator` and `TimeRangePaginator`, so I made the initial paginator an abstract class called `PaginatorAbstraction`, did not export it (so it is not available outside of the file), and then exported implementations:

```ts
export class Paginator<T> extends PaginatorAbstraction<T, PaginationParams> {}

export class TimeRangePaginator<T> extends PaginatorAbstraction<T, TimeRangePaginationParams> {}
```

If you think that just having the duplicate implementations is fine, I can remove the abstract class and just have the same code for `Paginator` and `TimeRangePaginator` with the types replaced! Trade-offs of abstraction vs. readability 😄 

(Also I am open to suggestions of a better name for `PaginationAbstraction`!)

ajaygandecha:
If we want to simplify `Paginated<T, Params>` to just `Paginated<T>` as well, we can by creating two separate pagination objects, like `Paginated<T>` and `TimeRangePaginated<T>`.

ajaygandecha:
Will merge, but further iteration on this will be done!

Issue #459, opened 2024-05-23
Refactor Events Backend Based on Frontend Refactor


ajaygandecha:
This ticket adds a to-do list on some possible improvements to the events backend as I refactor the frontend:

- [ ] Ensure that registration data does not emit all user details - should only emit profile name. Then, create a pagination API route for `EventRegistration[]` rather than just `User[]`, and replace `getRegisteredUsersForEvent` with `getEventRegistrations` on the frontend.
- [ ] Rename the register for event API to be a post as follows: `/api/events/register`. The body should be the event ID to register for. Same for unregistering.
- [ ] Refactor the events pagination such that by default, filtering by string does not use the time parameters.

ajaygandecha:
@jadekeegan pinging since we are both mentioned on this ticket -- will be closing, this will roll over into #482

PR #462, opened 2024-05-25
Angular v17 Upgrade: Events Feature Refactor

ajaygandecha:
This major pull request refactors the entire events feature to the new standards of Angular v17. This pull request also rewrites most of the functionality to be more concise and readable, following the best conventions for future students in COMP 423 and 393, using everything we have learned so far.

This refactor also fully utilizes the new frontend pagination abstraction created in #456.

*Once again, I am adding @jadekeegan and @atoneyd to this PR as reviewers, since this was the feature we worked on the most! @atoneyd, you can also see the first glimpses of the new age of pagination in our site - thank you for helping to get the first version working, which was the stepping stone that allowed the abstraction to be made!!✨ @jadekeegan, this PR also iterates on event registrations.*

### Intro

This refactor was the first one that featured pagination as the main form of data retrieval, marking the first usage of the `Paginator<T>` abstraction introduced in #456. This is a refactor using Angular v17. We can combine the signals functionality with the abstracted paginator to make our code much more readable and performant.

### The Event Service

To use our abstraction, the new event service creates an encapsulated event paginator object, as shown below:

https://github.com/unc-csxl/csxl.unc.edu/blob/fcf9e16ced9da94d81ac41cefe3e622456117dbc/frontend/src/app/event/event.service.ts#L34-L36

The `TimeRangePaginator` is an implementation of the `PaginatorAbstraction` abstract class which specifically works with paginated data on a time range, where a *start* and *end* time are used as parameters.

Loading our new pages is extremely easy and is done in the `getEvents()` method, shown below:

https://github.com/unc-csxl/csxl.unc.edu/blob/fcf9e16ced9da94d81ac41cefe3e622456117dbc/frontend/src/app/event/event.service.ts#L47-L49

This method takes in an optional `params` field for the pagination's parameters. In the case that this is not supplied, a default time range parameter constant is used. This saves us a lot of repeated code elsewhere in the feature.

As mentioned in #456, `.loadPage()` takes in both *params* and an operator function in the case that the model returned by the API is not of the same type expected. Since our API returns `EventJson`, we can pass in `parseEventJson` into this parameter, resulting in paginated data of type `Event`.

Notice that, unlike the organization refactor ( #455 ), the service does not expose any signals. This is because, unlike with organizations, events does not have a single collection of events shared by every component. We only want to load the events we need for a page (the entire point of pagination), so sharing this data does not really make sense here. Not to mention, pagination parameters are component specific and based on query parameters. So, this management is left up to the components.

The rest of the service methods follow normal conventions and return observables.

### The Event Page Component

Pagination truly shines in the event page component, which got the heaviest refactor - this component shrunk nearly in half after the refactor.

First, our component has a reactive `page` signal that stores the current paginated event data. This signal will update every time a new page is loaded to store the new page, allowing our page to dynamically update.

In addition, we previously needed to use a function to group event data by date. This functionality was moved into a new pipe called `GroupEventsPipe`. Now, the component also adds a *computed signal* which, when called (and the page updates), the signal uses the new pipe to expose grouped event data. This is done as shown below:

https://github.com/unc-csxl/csxl.unc.edu/blob/fcf9e16ced9da94d81ac41cefe3e622456117dbc/frontend/src/app/event/event-page/event-page.component.ts#L51-L55

The best part is that we can make our specific pagination parameters as *signals*. There are three parameters - one for the start date, end date, and any filter query. This is shown below:

https://github.com/unc-csxl/csxl.unc.edu/blob/fcf9e16ced9da94d81ac41cefe3e622456117dbc/frontend/src/app/event/event-page/event-page.component.ts#L56-L62

This functionality is great paired with an *effect*. The new component has an effect such that, when any of the pagination signals are updated, the page automatically refreshes with new data and the query parameters in the URL is updated:

https://github.com/unc-csxl/csxl.unc.edu/blob/fcf9e16ced9da94d81ac41cefe3e622456117dbc/frontend/src/app/event/event-page/event-page.component.ts#L81-L100

This is great, since our page now updates whenever a query parameter is changed - which is done when the search bar is edited, or when a user presses on the arrows to switch between pages.

### The Event Detail Component

Nothing major changed in the event detail component's refactor, although some code has been cleaned up. This component still uses the resolver to load a single event. This is required for the `titleResolver` to run properly.

### The Event Editor Component

The event editor has also been refactored to make the code cleaner and more concise.

In this refactor, the authentication functionality was moved to a new guard. Since this guard uses two different conditions to ensure that the user can access the page, I had to use the `combineLatest` RxJS method:

https://github.com/unc-csxl/csxl.unc.edu/blob/fcf9e16ced9da94d81ac41cefe3e622456117dbc/frontend/src/app/event/event-editor/event-editor.guard.ts#L45-L50

Notice though that this guard has to call the `getEvent()` function from the service - which repeats functionality in the event resolver. This means that when we open the editor, two GET API calls are performed. ***This is really unfortunate***, and I tried many different approaches, none of them work.

The problem is that resolvers run before `canActivate` guards in Angular, so there was no way to use the resolver's `event` data in the guard. The only other idea I saw online was to perform a reroute in the resolver if the user is not an organizer, however I felt that this overloaded the resolver with functionality that it should not have - thoughts @KrisJordan?

### Widgets

All widget HTML code was also updated to the new conventions! 

### Removal of the Event Admin Page and Gears

Throughout the refactoring process, I found the event admin component and related functionality merged in #402 needs to be reworked - so, I removed the admin page entirely. We will revisit this page once we get to the Angular Material 3 upgrade. No functionality is lost to the user with this removed, and the code becomes much cleaner.

### Reversion to a Single Column Format

I have temporarily disabled the dual-column arrangement on the events page, since you recommended this @KrisJordan (@jadekeegan is very sad 😭)

I mainly disabled it because the code became much more concise - even if we add it back, I think the way it was implemented needs reworking.

### File Structure Changes

There are not extremely major change to the file structure in the events page, however a new `/pipe` directory has been added to follow the conventions established in #455. 

## Future Considerations
* There are a few things I noticed in the events frontend that I think could warrant some refinements to our backend. These are noted in #459.
* We might be able to try to figure out a better construction to replace the existing `titleResolver` and the child route added to get it configured.
* As @KrisJordan mentioned, it might be better for all admin features to be moved to a separate admin folder rather than individual feature folders. Since services are not declared in modules and are injected from root, feature services would be available to the Admin module. This might allow for our file organization to be cleaner.


frontend/src/app/pagination.ts
```diff
@@ -31,6 +31,16 @@ export interface TimeRangePaginationParams extends URLSearchParams {
...
 }
 
+export const DEFAULT_TIME_RANGE_PARAMS = {
+  order_by: 'time',
+  ascending: 'true',
+  filter: '',
+  range_start: new Date().toLocaleString('en-GB'),
+  range_end: new Date(
+    new Date().setMonth(new Date().getMonth() + 1)
+  ).toLocaleString('en-GB')
```
KrisJordan, line 41: I'm still a bit perplexed by the need to use Great Britain English as the locale string. It looks like the difference is ordering of month/day/year vs day/month/year, as well as military time vs. AM/PM.

~~~plaintext
new Date().toLocaleString("en-US")
> '5/26/2024, 12:36:51 PM'
new Date().toLocaleString("en-GB")
> '26/05/2024, 12:36:53'
~~~

My feeling is the API should be defaulting to ISO style strings:

~~~plaintext
new Date().toISOString()
'2024-05-26T16:37:56.630Z'
~~~

This format is pretty standard and timezone independent (the Z indicates it's been converted to Greenwich Standard Time +0000). Without this, it seems we are relying on the timezone of the client and server both being in agreement. This is _generally_ true, but not universally, and it will lead to surprising errors when these are out of agreement.

I'm going to create a ticket that references this PR as motivation for moving date endpoints to standardizing on ISO format.

frontend/src/app/pagination.ts
```diff
@@ -70,7 +80,7 @@ abstract class PaginatorAbstraction<T, Params extends URLSearchParams> {
    */
   constructor(
     protected api: string,
-    protected http: HttpClient
+    protected http: HttpClient = inject(HttpClient)
```
KrisJordan, line 83: Is this idiomatic and necessary in new Angular?

I thought the use case for `inject` was more to declare `http` as a local member (at the top of the class definition) and assigning it `inject(HttpClient)` there, to avoid over-reliance on constructor parameters declaring members.

KrisJordan:
```suggestion
    protected http: HttpClient = inject(HttpClient)
```

frontend/src/app/app.module.ts
```diff
@@ -42,6 +42,7 @@ import { AboutComponent } from './about/about.component';
 import { GateComponent } from './gate/gate.component';
 import { ProfileEditorComponent } from './profile/profile-editor/profile-editor.component';
 import { SharedModule } from './shared/shared.module';
+import { GroupEventsPipe } from './event/pipes/group-events.pipe';
```
KrisJordan, line 45: Should this be in `event.module.ts` instead?

frontend/src/app/profile/profile.service.ts
```diff
@@ -27,11 +27,11 @@ export interface Profile {
...
-  id: number;
-  first_name: string;
-  last_name: string;
-  pronouns: string;
-  email: string;
+  id: number | null;
+  first_name: string | null;
+  last_name: string | null;
+  pronouns: string | null;
+  email: string | null;
```
KrisJordan, line 34: Should these really be `| null`? It seems to me every `PublicProfile` should have values for each of these fields and it adds complexity everywhere they are used to check for `null`. The only field that seems arguably hideable is `email`.

frontend/src/app/event/event-page/event-page.component.ts
```diff
@@ -3,268 +3,156 @@
...
+   * @param endDate: The new end date
+   */
+  reloadQueryParams() {
     this.router.navigate([], {
       relativeTo: this.route,
       queryParams: {
-        start_date: this.startDate.toISOString(),
-        end_date: this.endDate.toISOString()
+        start_date: this.startDate().toISOString(),
+        end_date: this.endDate().toISOString()
```
KrisJordan, line 134: Here we're using `toISOString` in the query!

frontend/src/app/event/event-page/event-page.component.ts
```diff
@@ -3,268 +3,156 @@
...
-      paginationParams.filter = this.query;
-    }
-    this.eventService.list(paginationParams).subscribe((page) => {
-      this.eventsPerDay = this.eventService.groupEventsByDate(page.items);
-      paginationParams.filter = '';
+  paginationTimeRangeEffect = effect(() => {
+    // Update the parameters with the new date range
+    let params = this.previousParams;
+    params.range_start = this.startDate().toLocaleString('en-GB');
+    params.range_end = this.endDate().toLocaleString('en-GB');
```
KrisJordan, line 92: Here we're using the Great Britain locale... hmm

frontend/src/app/event/event-page/event-page.component.ts
```diff
@@ -3,268 +3,156 @@
...
+    this.startDate.set(
+      query === ''
+        ? new Date()
+        : new Date(new Date().setMonth(new Date().getMonth() + 1))
+    );
+    this.endDate.set(
+      query === ''
+        ? new Date()
+        : new Date(new Date().setMonth(new Date().getMonth() + 1))
+    );
```
KrisJordan, line 155: Slightly confused by what this logic is doing, perhaps a comment would help.

The ternary operator's results seem strange. If query is empty, set start and end to the current date time. Oterwise, if non-empty, set both to one month from now?

frontend/src/app/event/event-editor/event-editor.guard.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  // Load IDs from the route
+  let organizationId: string = route.params['orgid'];
+  let eventId: string = route.params['id'];
+
+  // Create two observables for each check
+
+  // Checks if the user has permissions to update events for
+  // the organization hosting this event
+  const permissionCheck = inject(PermissionService).check(
```
KrisJordan, line 35: Add a `$` suffix to the variable name to make it clear this is an abservable. Same with `isOrganizerCheck` below.

frontend/src/app/event/event-editor/event-editor.guard.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  // Checks if the user is the organizer for the event
+  const isOrganizerCheck = inject(EventService)
+    .getEvent(+eventId)
+    .pipe(map((event) => event?.is_organizer ?? false));
+
+  // Since only one check has to be true for the user to see the page,
+  // we combine the results of these observables into a single
+  // observable that returns true if either were true.
+  return combineLatest([permissionCheck, isOrganizerCheck]).pipe(
```
KrisJordan, line 48: Nice use of Rx!

frontend/src/app/organization/organization-admin/organization-admin.component.ts
```diff
@@ -36,25 +35,19 @@ export class OrganizationAdminComponent {
...
 
   constructor(
-    private route: ActivatedRoute,
     private router: Router,
     private snackBar: MatSnackBar,
+    private profileService: ProfileService,
     private organizationService: OrganizationService,
     private permissionService: PermissionService
   ) {
+    this.profile = this.profileService.profile()!;
```
KrisJordan, line 48: The use of `!` here feels unsatisfying. Opening this route incognito leads to an error in console of trying to read permissions of undefined.

Perhaps this page should be protected by a guard of some sort. The other thing this points out to me, though, is perhaps there's a better way of ensuring that for pages we expect a user to be logged in on, we find a cleaner, idiomatic way to get that context into the components and can avoid this kind of coercion.

frontend/src/app/organization/organization-details/organization-details.component.ts
```diff
@@ -25,6 +25,7 @@ import { EventService } from '../../event/event.service';
 import { Event } from '../../event/event.model';
 import { Observable } from 'rxjs';
 import { PermissionService } from '../../permission.service';
+import { GroupEventsPipe } from 'src/app/event/pipes/group-events.pipe';
```
KrisJordan, line 28: Here we're using an import path that isn't as relative as it should be.

frontend/src/app/organization/organization-details/organization-details.component.html
```diff
@@ -7,13 +7,13 @@
...
     [eventsPerDay]="eventsPerDay"
-    [organization]="organization()"
+    [organization]="organization"
     [selectedEvent]="null"
     [disableLinks]="false"
     [showHeader]="true"
-    [showCreateButton]="(eventCreationPermission$ | async)!" /> -->
-</div>
+    [showCreateButton]="(eventCreationPermission$ | async)!" />
+</div> -->
```
KrisJordan, line 18: Should we just delete this dead code?

frontend/src/app/organization/organization-page/organization-page.component.ts
```diff
@@ -27,8 +26,7 @@ export class OrganizationPageComponent {
     path: '',
     title: 'CS Organizations',
     component: OrganizationPageComponent,
-    canActivate: [],
-    resolve: { profile: profileResolver }
+    canActivate: []
```
KrisJordan, line 29: This can be removed, yea?

KrisJordan (changes requested):
Re: The problem is that resolvers run before canActivate guards in Angular, so there was no way to use the resolver's event data in the guard.

Did you try accessing the `data` of `route` to access the resolved data here: https://github.com/unc-csxl/csxl.unc.edu/blob/fcf9e16ced9da94d81ac41cefe3e622456117dbc/frontend/src/app/event/event-editor/event-editor.guard.ts#L24

Looking great! Pointed out a few places where I think things could be tidied or some open questions existed that are worth moving through before merging in.

frontend/src/app/event/event-page/event-page.component.ts
ajaygandecha: I agree, I think it would be better to use a regular if-statement here! This ended up like this due to the original code and also a few things being moved around in the refractor.

ajaygandecha: Yeah @KrisJordan, the logic here seems broken - I think it was as follows (and how I just edited it to be):
* If the query is blank, set back to a range of today through today + 1 month
* If the query is not blank, set the range to be (-100 years, +100 years)

I don't necessarily think that this is good, but it is what we had before! I am thinking this rate range should be handled in the backend when a filter query is provided. Thoughts?

frontend/src/app/profile/profile.service.ts
ajaygandecha: Agree!

frontend/src/app/pagination.ts
ajaygandecha: @KrisJordan in this case, removing `injected(HttpClient)` causes an error:
```
An argument for 'http' was not provided.
```
I think since this is in a standalone class, Angular does not seem to automatically inject the HTTPClient. So, in order to prevent having to pass the injected `HTTPClient` instance into each paginator object in every service, this small addition directly injects the client into the object.

Now, it might be better to move this out of the constructor and have it as a field - thoughts?

ajaygandecha: (Also, the `;` suggestion causes a syntax error since it is in the constructor, so I did not include!)

ajaygandecha:
Thank you for the review @KrisJordan! I fixed most of the comments, but I had some additional comments on the rest + some other things that I have noticed. 

1. Yes, the date conversions here were extremely painstaking - in the first iteration of COMP 393, we ended up having to use the `GB` version to get everything working. We had issues with getting the native date selectors to work, as they reverted back to random / arbitrary times. I think, however, that the solution separating the Json model and the regular model may have solved the date problems we had - so, I will revisit this now and see if I can get things working standardized on ISO format!

2. For the resolver / guard issue, my first attempt actually was to use `route.data` - however, the result was `undefined`. Since resolver data is not loaded before the guard runs, the route's data is not set by the time the guard runs. Now, I think that this problem also kind of ties into the comment you made about `profile()!`, something I also seems off. In this case too, even if we protect the admin page with a guard, it would still be hard to pass this data directly in. We could have a profile resolver which loads a profile if logged in, or redirects the user if they are not logged in, but it also seems really anti-Angular to me. We could create some sort of guard that is loaded in the app module and then pass it around, but not sure what is best here.
 
3. This is not related to the CR comments, but something I wanted to ask about as we did this refactor. There does not yet seem to be a convention marking signal variables like we have `$` for observables. In Swift state variables, we also use `$`, but preceding the name (as in `$flashcards`). Do you think we need to standardize on some convention to denote signals vs. non-signals? I can see, for example, `organizations` being more easily confused as a non-signal variable.

ajaygandecha:
Well, it looks like the conversion to ISO strings was easier than expected!

frontend/src/app/pagination.ts
KrisJordan: Ah, I see, right this is outside of dependency injection scope. Makes sense!

frontend/src/app/event/event-page/event-page.component.ts
KrisJordan: Got it, makes sense! Ultimately you could imagine wanting these two dimensions to be orthogonal and composable, not either-or (search for this term within this date range) but that becomes a UI problem (and it sounds like this is actually possible via the API logic described).

backend/services/event.py
```diff
@@ -77,9 +77,8 @@ def get_paginated_events(
             range_start = pagination_params.range_start
             range_end = pagination_params.range_end
             criteria = and_(
-                EventEntity.time
-                >= datetime.strptime(range_start, "%d/%m/%Y, %H:%M:%S"),
-                EventEntity.time <= datetime.strptime(range_end, "%d/%m/%Y, %H:%M:%S"),
+                EventEntity.time >= datetime.fromisoformat(range_start),
```
KrisJordan, line 80: Nice! Standards for the win!

KrisJordan (approved):
LGTM!

PR #464, opened 2024-05-25
Fix Critical Issue - Applications Data Leaking with Section Data

ajaygandecha:
This pull request removes the `preferred_applicants` field from the backend section model. This field leaks users' application data via the public GET section data API. In addition, this data caused pages relying on sections data to load extremely slowly.

After further analysis, it does not appear that this extra field had any functionality in the backend, so I think it can be safely removed.


KrisJordan (approved):
LGTM

PR #497, opened 2024-06-08
Index Academics join table columns that back common queries

KrisJordan:
By adding Indexes to Academics' SectionMemberEntity (by user and by section), and SectionEntity (by term and by course), we should see an automatic performance increase for some common queries (e.g. look up all members of a section or all sections for a user).

This PR also drops the `application_id` column from `SectionMemberEntity` as a cleanup while we're running a migration on the table. When the time comes, this information (linking a section member to their TA application) can be held in a different entity, for example a `TAAssignmentEntity` which can join an application to a member (with additional information such as hiring level, and so on).

Migration created and validated per the `backend/migrations/README` validation steps.

### DIFF
--- a/backend/entities/academics/section_entity.py
+++ b/backend/entities/academics/section_entity.py
@@ -1,7 +1,7 @@
 """Definition of SQLAlchemy table-backed object mapping entity for Course Sections."""
 
 from typing import Self
-from sqlalchemy import Integer, String, ForeignKey
+from sqlalchemy import Integer, String, ForeignKey, Index
 from sqlalchemy.orm import Mapped, mapped_column, relationship
 from ..entity_base import EntityBase
 from ..section_application_table import section_application_table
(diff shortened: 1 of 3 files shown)


ajaygandecha (approved):
LGTM! All features work on localhost, and all tests pass. I think this is ready to merge!

@KrisJordan for a knowledge thing,I am curious though, how does defining these as indices differ from establishing a primary key with these columns? Is it that we are adding another way for a quick indexing/lookup while also keeping a separate unique identifier for each column?

KrisJordan:
Yes, precisely: adding secondary indexes leads to the rdbms maintaining a separate B-tree (or comparable) per index. Primary keys will automatically be indexed.

When a secondary index has multiple columns, you get the lookup benefits when looking up by either the first column in the index or both columns.

Without any indexes to rely on, searches will linearly scan results.

The trade-off is slower writes and some space, but for our join tables and typical cases, going from o(n) reads to o(log n) is worth the win (esp for joins where you start trending toward o(n^2) without an index to rely on).

PR #500, opened 2024-06-14
Create My Courses Homepage and the Roster View

ajaygandecha:
This incremental pull request covers the creation of the My Courses homepage component and the roster component, as well as related service and API features.

## Major Changes

* Adds the my courses page at the route `/my-courses`, which shows the courses the user is enrolled in (or manages).
* Adds a tab page for each course at `/course/:termid/:courseid`, which contains three tabs - one to contain the future office hours page, one that shows the course's roster, and another for settings. Office hour and settings still has to be implemented, and full security for these routes also has to be implemented.
* Adds the roster page at `/course/:termid/:courseid/roster`, which contains a paginated table showing members of a course.
* Adds an API at the route `/api/academics/my-courses/:termid/:courseid/roster` which returns a paginated list of users in a course. Only instructors and students can access this data.
* Adds related models and service methods for the newly added API.
* Moves academics pages into a new catalog page with tabs separating current section offerings and all courses.


ajaygandecha:
@KrisJordan just wanted to check in, mainly with the roster service method, to ensure that the query is built in a way that is not too inefficient!

backend/services/academics/my_courses.py
```diff
@@ -99,4 +104,109 @@ def _to_section_overview(self, section: SectionEntity) -> SectionOverview:
...
+        # Start building the query
+        member_query = (
+            select(SectionMemberEntity)
+            .join(SectionEntity)
+            .join(UserEntity)
+            .where(
+                SectionEntity.term_id == term_id,
+                SectionEntity.course_id == course_id,
+            )
+            .options(joinedload(SectionMemberEntity.user))
```
KrisJordan, line 133: Is there an intentional reason not to do the `joinedload` of `SectionMemberEntity.section` here, in addition to `.user`? I can imagine some potential reasons, but I would personally still default to going ahead and eagerly loading it in the one query rather than having SQLAlchemy lazily load it in one or more follow-on queries. That said, if you read documentation that would suggest letting an oft-repeated related entity be loaded lazily, I would defer to it.

backend/services/academics/my_courses.py
```diff
@@ -99,4 +104,109 @@ def _to_section_overview(self, section: SectionEntity) -> SectionOverview:
...
+
+        # Determines if a user is a student
+        # NOTE: This can be used to limit roster data a user can see compared to
+        # an instructor in the future.
+        is_student = user_members[0].member_role == RosterRole.STUDENT
+
+        # In the cases where sections are taught by different instructors, ensure that
+        # the roster data only includes sections that the user has permissions for.
+        section_ids = [member.section_id for member in user_members]
+        member_query = member_query.where(SectionEntity.id.in_(section_ids))
```
KrisJordan, line 161: 👍

backend/services/academics/my_courses.py
```diff
@@ -99,4 +104,109 @@ def _to_section_overview(self, section: SectionEntity) -> SectionOverview:
...
+
+        # Count the number of rows before applying pagination and filter.
+        count_query = select(func.count()).select_from(member_query.subquery())
+        length = self._session.scalar(count_query)
```
KrisJordan, line 165: 🔥

KrisJordan (approved):
This looks SO GOOD, wow. The design really came together for the home page and I am loving the top tab bar on the individual course. Nice work getting the roster functionality paginated, too. This seems ready to merge into the dev branch for me.

PR #501, opened 2024-06-17
Implement Office Hours Feature for the My Courses Page

ajaygandecha:
This (large) incremental pull request covers the re-implementation of the Office Hours feature's backend services, APIs, and frontend.

## Demo Instructions

As mentioned in the "Next Steps" section below, there is not currently frontend functionality to manage office hour events - that will be in the next PR, since it is more involved. For now, testing has been done based on the test data. Amy Ambassador is listed as a GTA for COMP 110, and Sally Student is a student for COMP 110. Use these two members for testing out the queue process!

### The Frontend
* The old office hours frontend has been removed and replaced with a new set of components at the path `my-courses/course/office-hours`.
* Added three components to handle the office hours feature. Between these pages, the entire office hours functionality is encapsulated:
    * **Office Hours Page**: This component is the landing page when accessing a course. It shows the currently active office hour events, as well as a paginated preview of the future events. There is also another tab showing all past office hour events, and all data is paginated.
    * **Office Hours Queue Page**: The queue page is accessible to UTAs, GTAs, and instructors for a course. This page shows the active queue and called tickets. Students are unable to access this page. This page updates every 10 seconds with the latest overview of the course's office hours queue.
    * **Office Hours Get Help Page**: The get help page is accessible to students of a course. If a students has an active ticket, the ticket will show, as well as the place in the queue (or, show if the ticket has been called). If the student does not have an active ticket, this page will show the ticket creation form. When the user submits a ticket, the page's will automatically switch to the page showing its progress in the queue. This page also updates every 10 seconds with the latest overview of the getting help progress.

### The Backend
* The old office hours APIs and services have been removed and replaced with APIs and services within `my_courses`. Some old models have moved to the `my_courses` models.
* Created the following APIs for office hours:
    Screenshot 2024-06-17 at 4 32 29 PM
* All of these APIs return `Overview` models that are tailored to the data required by the frontend. No extra data is passed along to the frontend except what is needed.

## Next Steps

* Pytest unit testing for these new services has not yet been completed and will happen in the next pull request.
* The only frontend functionality missing right now is for the *creation and managing of office hour events and section members*. This requires more discussion will happen in the next PR, **so for now, testing can be done by authorizing as `Sally Student` and `Amy Ambassador`, and attempting to interact with `COMP 110 Office Hours` as defined in the testing data.**
* Ideally, the office hours functionality of `my-courses` will be re-refactored out, but to avoid confusion in the PR history with old office hours files, this will be done in the next PR.
* The title resolver for the courses tab page still has to be implemented.
* All office hours features besides the office hours page might be moved out of the tab bar page to avoid confusion. For example, the queue would not have the tab bar saying "Office Hours | Roster | Settings" above it. Thoughts @KrisJordan?


backend/services/academics/my_courses.py
```diff
@@ -207,6 +225,639 @@ def _to_course_member_overview(
...
+        Gets the past office hours events, paginated.
+
+        Returns:
+            Paginated[CourseOfficeHourEventOverview]
+        """
+
+        # Start building the query
+        event_query = self._create_oh_event_query(user, term_id, course_id)
+
+        # Only load future events
```
ajaygandecha, line 325: Typo: past

ajaygandecha:
Ah - nighttime thought! Deleted the test data file, which might cause problems. Will revert tomorrow

ajaygandecha:
@KrisJordan I reverted the latest few commits that deleted the old models / tests of the old office hours feature, since I realized that the new feature is dependent on the test data from the old feature. Once this PR is approved, the next PR will be a neater refactor of all of the features that were added here, as well as testing and a cleanup of the old models!

ajaygandecha:
@KrisJordan I have worked a bit more on this and completed some of the future task items - including refactoring out the newly-created APIs to be organized, separating `api/my-courses` from `api/office-hours`. I decided to opt for office hour events based on term and course to be an API under My Courses, since that seems to be its main use case.

The new API organization is summarized below:

Screenshot 2024-06-20 at 12 37 43 AM

I also refactored the services and APIs to be organized this way. The first iteration of the solution put all of the office hours APIs into `api/my-courses.py`. Now, they are neatly organized, and so are the services.

The only thing that has gotten messier is the `_to_..._overview()` methods. I am not sure what the best place for this is. I wonder if we want to put it back in the entity, or move away from the entity convention entirely, create some sort of bridge file containing such model conversions, or just use constructor overloading in the model file (if Python allows that feature).

Another thing I noticed while refactoring was the sheer number of `academics` APIs. I decided to remove a few of the unused ones - but, there are still a lot:

Screenshot 2024-06-20 at 12 41 19 AM

We can probably prune further. I would also not be opposed to moving some of the APIs into APIs at routes like `/api/my-courses/catalog`.

(Also, while we are at it, `GET /api/sections` should probably just be removed - it loads data for every single term at once, a very expensive query)

backend/api/office_hours/event.py
```diff
@@ -1,287 +1,66 @@
...
+from ...models.user import User
 
+from ...models.academics.my_courses import (
+    OfficeHourQueueOverview,
+    OfficeHourEventRoleOverview,
+    OfficeHourGetHelpOverview,
+)
 
-__authors__ = ["Sadie Amato", "Madelyn Andrews", "Bailey DeSouza", "Meghan Sun"]
+__authors__ = ["Ajay Gandecha"]
```
KrisJordan, line 17: We should leave attribution here even though the rewrite is significant; not starting from a position of scratch was thanks to the initial effort of the original authors. Additionally, they'll be in the git log with blame, etc.

backend/api/office_hours/ticket.py
```diff
@@ -1,201 +1,78 @@
...
 from ..authentication import registered_user
-from ...models import User
+from ...services.office_hours.ticket import OfficeHourTicketService
+from ...models.user import User
+from ...models.office_hours.ticket import OfficeHoursTicketDraft
 
+from ...models.academics.my_courses import OfficeHourTicketOverview
 
-__authors__ = ["Sadie Amato", "Bailey DeSouza", "Meghan Sun"]
+__authors__ = ["Ajay Gandecha"]
```
KrisJordan, line 14: Same here

backend/models/office_hours/event_type.py
```diff
@@ -12,16 +12,16 @@ class OfficeHoursEventType(Enum):
     Determines the type of an office hours event.
     """
 
-    OFFICE_HOURS = 0
-    TUTORING = 1
-    REVIEW_SESSION = 2
+    OFFICE_HOURS = "Office Hours"
+    TUTORING = "Tutoring"
+    REVIEW_SESSION = "Review Session"
```
KrisJordan, line 17: I am slightly concerned this is going to cause a migration and one that will take some care to pull off (I would actually need to think through exactly how...) since we already have a column storing this data in an int form (I believe). The process of changing to a string/text and preserving the data from SS1/SS2 may be tricky. For now, maybe we should leave this as ints and make this something to come back to. Ultimately, I could see this being converted to an entity that courses are able to control and customize.

backend/models/office_hours/event_type.py
```diff
@@ -12,16 +12,16 @@ class OfficeHoursEventType(Enum):
...
     """
     Determines the office hours event mode.
     """
 
-    IN_PERSON = 0
-    VIRTUAL_STUDENT_LINK = 1
-    VIRTUAL_OUR_LINK = 2
+    IN_PERSON = "In-Person"
+    VIRTUAL_STUDENT_LINK = "Virtual - Student Link"
+    VIRTUAL_OUR_LINK = "Virtual - Our Link"
```
KrisJordan, line 27: Same concern here going from int to string. For now, I'd say let's revert. Perhaps at the model layer we can convert to string with a custom function that does the quick conversion, if the model is expecting a string for API endpoints.

backend/models/office_hours/ticket_state.py
```diff
@@ -12,7 +12,7 @@ class TicketState(Enum):
...
     """
 
-    QUEUED = 0
-    CALLED = 1
-    CLOSED = 2
-    CANCELED = 3
+    QUEUED = "Queued"
+    CALLED = "Called"
+    CLOSED = "Closed"
+    CANCELED = "Cancelled"
```
KrisJordan, line 18: Same with this conversion; revert to int.

backend/models/office_hours/ticket_type.py
```diff
@@ -12,5 +12,5 @@ class TicketType(Enum):
     Determines the type of a ticket.
     """
 
-    CONCEPTUAL_HELP = 0
-    ASSIGNMENT_HELP = 1
+    CONCEPTUAL_HELP = "Conceptual Help"
+    ASSIGNMENT_HELP = "Assignment Help"
```
KrisJordan, line 16: Same with this conversion; revert to int.

backend/models/roster_role.py
```diff
@@ -8,7 +8,7 @@
...
 
 class RosterRole(Enum):
-    STUDENT = 0
-    UTA = 1
-    GTA = 2
-    INSTRUCTOR = 3
+    STUDENT = "Student"
+    UTA = "UTA"
+    GTA = "GTA"
+    INSTRUCTOR = "Instructor"
```
KrisJordan, line 14: Same with this conversion; revert to int.

backend/services/academics/my_courses.py
```diff
@@ -81,7 +85,7 @@ def _group_by_term(self, entities: list[SectionMemberEntity]) -> list[TermOvervi
                         subject_code=course.subject_code,
                         number=course.number,
                         title=memberships[0].section.override_title or course.title,
-                        role=memberships[0].member_role.name,
+                        role=memberships[0].member_role.value,
```
KrisJordan, line 88: Let's use the names here (if we need pretty casing, perhaps use a conversion function)

KrisJordan (changes requested):
This is really coming together and looking really sick; great work! There are a few minor changes we should make in this PR before merging and then this comment alludes to some larger refactorings I think we should go ahead and prioritize so that we don't have more backtracking to do after. 

A potentially significant issue with the URL system for a route like (/course/SSI24/comp110/office-hours): not all sections of a course for a given term are going to share the same team/OH. For example, 110 may have two different teams with two instructors this Fall; in the past when two instructors have taught the same section (thinking of 311 last Fall) they had different teams/queues/rosters as well.

Additionally, when thinking about how Settings will be stored, it seems like we'll need a new entity to store configurable settings of a course.

This is all making me think we really do need an entity (and some UI for setting it up) for an "Offered" or "Instructor's" Course (that can have multiple sections of a given instructor associated with it) separate from a "Catalog" Course (that has all sections associated with it). I could potentially see a rationale for renaming our current "Course" entity `CatalogCourse` and giving the notion of an offered/instructor's Course the plain `Course` moniker. This is effectively the role `OfficeHoursSectionEntity` was playing, so perhaps we refactor rename this and use it since it's already doing most everything we want this entity to do; but it is more pervasive than just office hours and "section" is kind of the wrong term as multiple sections will relate to it; `Course` feels right to me here since we could imagine a lot of functionality (beyond OH) being built for specific instances here (Canvas should be scared...).

I could imagine a greenfield "setup" flow being:

1. Instructor goes to My Courses and is presented with a button to "Create Course" (or more ideally, but more complexity, shows sections they are the instructor of, but not associated with an instructor's `Course` yet, and a "Create Course" button is shown there).
2. Instructor selects the section (singular) they want to create a course for.
3. Instructor is taken directly to the Settings page of the course and one of the Settings is "Sections" associated with the course. From here, they can add additional sections (if any).

Upon viewing My Courses after associated with an instructor's `Course`, the Courses are shown the same as they are now.

A few nits clicking through the UI:

1. From the *My Courses* page, do we anticipate more than one button for students and or staff/instructors? Or is it always jumping straight to the course page? If so, should the card itself be hoverable/clickable without a button?
2. The student's view of My Courses should probably say "View" or "Office Hours" if we're keeping the button.

backend/api/office_hours/event.py
ajaygandecha: @KrisJordan sorry! I think this was caused because I deleted the original file and created a new one with the same name, causing this overwrite. Will revert 🫡

ajaygandecha:
@KrisJordan I think that refactoring the entities is a very good idea! I refrained from changing entities too much in this PR, but it definitely became evident that something felt off. Like the new suggestions and I will see if I can get a diagram of this done soon!

> A potentially significant issue with the URL system for a route like (/course/SSI24/comp110/office-hours): not all sections of a course for a given term are going to share the same team/OH. For example, 110 may have two different teams with two instructors this Fall; in the past when two instructors have taught the same section (thinking of 311 last Fall) they had different teams/queues/rosters as well.

This was remedied by backend APIs cross-listing the sections that the user was in and returning the relevant results. Still though, agree that this would be a problem for sharing. I think that the main problem is the relationship between courses and sections (something the entity refactor can achieve). 

In general though, I feel like the main question I have regards the administration side of courses, instructors, sections, and office hours. The current solution, as well as the newer solution, seems a bit complex.

ajaygandecha:
https://github.com/unc-csxl/csxl.unc.edu/blob/ce6136b5376db87086f329d5d26f4d8bc990ba00/frontend/src/app/my-courses/widgets/course-card/course-card.widget.html#L23-L29

Make buttons `<a>` tags

ajaygandecha:
> https://github.com/unc-csxl/csxl.unc.edu/blob/ce6136b5376db87086f329d5d26f4d8bc990ba00/frontend/src/app/my-courses/widgets/course-card/course-card.widget.html#L23-L29
> 
> Make buttons `<a>` tags

Making a new issue for this, since this will be a problem site-wide: #502

ajaygandecha:
@KrisJordan, the next CR is ready! 👀 

## Changes from the Previous CR
* Renamed `OfficeHoursSectionEntity` to be `CourseSiteEntity` and add a relationship with the term table.
* Renamed `OfficeHoursEventEntity` to be `OfficeHoursEntity`.
* Removed the `date` field from `OfficeHoursEntity`, since this information is redundant from `start_time` and `end_time`.
* Removed many unused / unnecessary models from the first office hours implementation, refactored all models to match the new entities.
* Rewrote all of the office hours test data. Now for demoing, instead of `Amy Ambassador`, `Ina Instructor` will be the instructor for COMP 110! The code is also shorter.
* Deleted all previously-implemented office hours testing (since those services were removed)
* Refactor `MyCoursesService` to become `CourseSiteService`, refactor all methods to use the new `CourseSite` models. Also uses the course site ID, rather than a combination of term ID and course ID, to find roster and office hour information.
* Add a few test cases. We are at 100% passing, but not 100% coverage yet. Will seek to do this in the next branch along with the create/update/delete features for course sites and office hour events.
* Modify the frontend to work with the backend changes. Routes now use course sites. Course sites now only show up on the "My Courses" page is a site has been made for them.

ajaygandecha:
@KrisJordan Update: Reached full coverage on all service methods added so far

backend/migrations/versions/1d7ab0cd0f58_migration_for_office_hours.py
KrisJordan, line 1: We'll want to revert these changes because migrations are applied linearly to the stateful database schema running behind the scenes. Once a migration is successfully run, it isn't run again (reason to potentially start clearing old ones out of the code base at some point, but they're decent historical artifacts that will coincide with certain commits in the history of the repo).

We'll ultimately need to make a new migration that modifies the existing schema by renaming columns and tables. I can take a stab at this when the time comes to deploy.

backend/models/office_hours/event_type.py
KrisJordan: I think this concern still holds here for these enums.

backend/services/office_hours/ticket.py
```diff
@@ -1,689 +1,280 @@
...
-        # Fetch Ticket By ID
-        ticket_entity = self._session.get(OfficeHoursTicketEntity, oh_ticket_id)
-
-        if ticket_entity is None:
-            raise ResourceNotFoundException(
-                f"Office Hours Ticket with id={oh_ticket_id} not found."
+        # If the user is not a member of the looked up course, throw an error
+        if not user_member or user_member.member_role == RosterRole.STUDENT:
+            raise CoursePermissionException(
+                "Not allowed to call if a ticket if you are not a UTA, GTA, or instructor for."
```
KrisJordan, line 102: In the current state of the queue, where it's updated based on polling, we'll want to be very careful about how this exception is handled client side: it's more common than you'd think, esp in a course like 110 with 5+ TAs on staff, that two people will call the same ticket within a 10s refresh window. When this happens, the person who called second should have some kind of reasonable UX in not getting an error but a meaningful message that the ticket was already called.

backend/services/office_hours/ticket.py
```diff
@@ -1,689 +1,280 @@
...
-            .filter(
-                SectionMemberEntity.section_id == SectionEntity.id,
-                SectionMemberEntity.member_role == RosterRole.STUDENT,
-                SectionMemberEntity.user_id.in_(ticket_creator_ids),
-            )
-            .distinct()
-            .all()
-        )
+        # Commit so can get ticket id
+        self._session.commit()
```
KrisJordan, line 264: I believe we should be able to do this all in a single transaction without committing first, then adding later. SQLAlchemy should be able to sync this.

backend/services/office_hours/ticket.py
```diff
@@ -1,689 +1,280 @@
...
-        # If the student currently has a queued ticket, don't allow new ticket creation
-        if (
-            self._oh_event_svc.check_student_in_queue_status(
-                subject, oh_event
-            ).ticket_id
-            != None
-        ):
-            raise PermissionError(
-                "Cannot create another ticket while currently in the queue."
-            )
```
KrisJordan, line 669: Is this code moved somewhere else or where is the logic for rate limiting ticket creation on a student's part?

KrisJordan (commented):
Getting close; a few observations and comments in the line or file specific reviews!

Biggest changes I think still needed:

1. Using ints for the enums (since we started that way and they're in the database)
2. Reverting the database migration changes (we need to make a new migration that renames existing things, doesn't try to recreate them). This will only block merging into prod when the time comes, but we can get a jump once this merges into the WIP branch.
3. Questino about office hours ticket rate limiting

backend/services/office_hours/ticket.py
ajaygandecha: @KrisJordan removing the commit does fail - I think this is because the `id` field is not filled in with any value until it has been persisted in the db.

backend/services/office_hours/ticket.py
ajaygandecha: @KrisJordan nope, re-added this check - thanks for catching this!

ajaygandecha:
@KrisJordan done! Let me know if the temporary version to convert enums <-> string representation looks good

ajaygandecha:
Also, the new query that determines whether or not a ticket is already in queue can be extended based on provided OH settings in the future - for example, if COMP 423 establishes a 30 min cooldown time, we can do something like `closed_at + timedelta(minutes=30) >= datetime.now()` (of course w/ more nuance for if `closed_at` is `None`)

backend/models/office_hours/event_type.py
```diff
@@ -16,6 +16,23 @@ class OfficeHoursEventType(Enum):
...
+            return OfficeHoursEventType.REVIEW_SESSION
+
+    def to_string(self) -> str:
+        if self == OfficeHoursEventType.OFFICE_HOURS:
+            return "Office Hours"
+        if self == OfficeHoursEventType.TUTORING:
+            return "Tutoring"
+        if self == OfficeHoursEventType.REVIEW_SESSION:
+            return "Review Session"
+
```
KrisJordan, line 35: Looks good for nicer strings than the `CONSTANT` style.

KrisJordan (approved):
This looks good to merge into feature branch.

PR #512, opened 2024-07-03
Implement Backend for Hiring Feature

ajaygandecha:
This incremental pull request implements the backend for the hiring feature, which allows instructors on course sites to view applicants and change their preference.

## Major Changes
* Added the `ApplicationReviewEntity` which connects course sites and applications.
* Create many overview-style models to return only required data to the frontend.
* Created the `HiringService` with the following functionality:
  * **Get hiring status:** Retrieves the hiring status that will be displayed on the hiring frontend, separating applications into three columns (see the `HiringStatus` model). This will pair very nicely with Angular's built-in drag-and-drop system. If an application is not associated with an `ApplicationReview` when this method is called, a blank review will be created and persisted in the database.
  * **Update hiring status**: Takes in a `HiringStatus` object, mutated by the frontend, and modifies the `ApplicationReview` data in the backend to match. This includes changes to the status (column the application is in), the ordering (preference), and any notes provided.
* Added two APIs:
  * **GET** `/api/hiring/:course_site_id`: Gets the hiring status.
  * **PUT** `/api/hiring/:course_site_id`: Updates the hiring status.

## Testing
* `HiringService` has 100% passing tests with `pytest`, as well as 100% coverage.

## ‼️ Outstanding Issue

@KrisJordan, you will notice the API returns `null` to data specific to the `NewUTAApplicationEntity`. I have burrowed deep into the SQLAlchemy docs and have tried *numerous* combinations to get this data to load, but there is no success. The `SectionEntity` is related to `UTAAplicationEntity`, which is a superclass of `NewUTAApplicationEntity`. SQLAlchemy docs point to [polymorphic eager loading](https://docs.sqlalchemy.org/en/20/orm/queryguide/inheritance.html#applying-loader-options-to-the-subclasses-loaded-by-selectin-polymorphic), however I have tried many permutations and combinations of what they recommend in the docs, and the Python debugger still notes that the data type of the section's application objects are `UTAAplicationEntity`. I am a bit at a loss - might have to ask for some help on this one 😃

### DIFF
--- a/backend/api/hiring.py
+++ b/backend/api/hiring.py
@@ -1,0 +1,48 @@
+"""Hiring API
+
+Hiring routes are used for hiring based on TA Applications."""
+
+from fastapi import APIRouter, Depends
+
+from ..services import HiringService
+
+from ..models.application_review import HiringStatus
+
+from ..api.authentication import registered_user
+from ..models.user import User
+
+__authors__ = ["Ajay Gandecha"]
+__copyright__ = "Copyright 2024"
+__license__ = "MIT"
+
+api = APIRouter(prefix="/api/hiring")
+
+openapi_tags = {
+    "name": "Hiring",
+    "description": "View and update the hiring status for a course site.",
+}
+
+
+@api.get("/{course_site_id}", tags=["Hiring"])
+def get_status(
+    course_site_id: int,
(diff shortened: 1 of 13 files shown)


KrisJordan (approved):
LGTM!

PR #529, opened 2024-07-18
Re-Implement the TA Applications Feature, Extend TA Applications Feature to Graduate TAs

ajaygandecha:
This major pull request re-implements the TA applications feature from the ground up, extending backend functionality to work for graduate teaching assistants. The frontend is also entirely redesigned to include the site's first dynamic form, creating forms based on the different types of applications without needing extra HTML.

### Major Changes
**Backend:**
* Removed polymorphic relationships between tables and instead rely on the `type` field to distinguish between application types. This makes the backend logic monumentally simpler and cleaner without having to worry about type casting and conflicts. 
* Simplified models.
* Rewrote the application services to include CRUD operations on multiple types of applications. All services now include test data and unit tests. All tests pass with 100% coverage for the new service.
* Simplified CRUD APIs.

**Frontend:**
* Created a dynamic form system for applications, enabling quick editing and extending of forms for different types of applications without ever needing to touch HTML. This also decreases the form's HTML file from 300 lines to ~19. These forms are stored inside of TypeScript objects in the new `application-forms.ts` file. *In the future @KrisJordan, we may be able to pull these form templates from the database / API directly.*
* Created the new application form component, service, and module to encapsulate all application features.
* The site currently supports application types of `gta` and `new_uta`, with easy extending in the future.
* To access an application, use the url `/apply/:term-id/:type`. For example, to apply as a GTA for the upcoming term, use `/apply/24F/gta`.
* The new forms component also enables both the creation of new applications and the editing of an existing one for the provided term, if it exists.
* Added a new, temporary banner to the home page for graduate applications.

### For Kris
This PR will require a migration to add new GTA-specific fields to the database.


ajaygandecha:
@KrisJordan (and @jadekeegan for an update), this PR now contains the new `markdown` directive to render markdown inside of HTML elements!
**Usage:**
```html
<p markdown>{{ exampleText }}</p>
```

KrisJordan (approved):
Really great work here! Having a flexible form solution is awesome; we'll have to get this going in the Fall for UTAs once the Spring course schedule is known.

One question came up I'm curious of: what's the behavior of viewing the form after an application period ends from the submitters perspective? I noticed the course selection API endpoint is driven by the current application term, so if you were viewing your application after the fact I'm not convinced the display of course selections would still work (even if the field is disabled)? I didn't investigate further, so perhaps this is a solved problem, but if not we may want to add it to a list of refactors for Fall going into Spring.

ajaygandecha:
@KrisJordan this is a good point! I *believe* that it should work since the section preferences are ultimately connected between an application ID and a section ID, so even after the current term, selections should be visible - however, this could cause problems if someone was trying to edit their application many months later with a new active term. I think the solution would be to disable editing for applications whose term window is closed (where the date is after the term's end date). We can add this as a ticket for the future!

PR #539, opened 2024-07-21
Implement the Welcome Page and News Articles Features

ajaygandecha:
This major pull request adds a new welcome page to the CSXL site that displays news articles and personalized information for users. The article feature enables the CSXL to post and edit articles, create announcements, and connect authors to articles. All pages are responsive and work on mobile.

## The Welcome Page

Screenshot 2024-07-21 at 1 32 03 AM

From the new welcome page, users are able to see any announcements that are active, as well as the most recent news articles. Users can also see the XL's operating hours, as well as any reservations or event registrations they have coming up. In the future, this page will be further customized to include user friendships and office hours information.

## Article Page

Screenshot 2024-07-21 at 1 32 39 AM

The article page renders markdown as shown in the Lorem Ipsum text, so article authors do not have to worry about custom HTML for every article page.

## Admin Features

This PR also includes full admin features, including creating / editing / deleting articles. At the moment, only the admin user or people with the `article` permission are able to add articles. This will be fine for our use-case.

## Backend

This pull request creates a new database table called `article` and an association table between `article` and `user` to keep track of authors. Due to this setup, an article can have many authors attached to it. This PR also creates new services, models, and API routes.

## Future Goals

* Enable organization leaders to submit articles for review.
* Create categories and tags for articles for filtering purposes.
* Allow users to access an article archive. Currently, only the top 10 articles are shown.


ajaygandecha:
@KrisJordan added an alternative view for people not signed in - the "For you" tile has been replaced by the about tile (also slightly improved!):

image

KrisJordan (commented):
This is looking really awesome. Great work bringing this together!

Some initial reactions in clicking around and thinking through:

1. On the latest news pane.
- Can we drop the modified date (or only show it if different from created)
- Let's remove the Read button and make the card clickable? Or make elements of the card clickable (title/icon/text) and take you to details view.
- I think we could lose the author line from home page and leave it just to the details page
- When a student org is listed, hovering should show hand click icon

2. Announcement.
- Reading an announcement details should not open a new tab
- The heading text feels too small relative to the heading text of a news post. Should we just use the same template as a news post but use the different highlighted color scheme?
- Should we 

3. Seeing some horizontal scrolling on the page at all widths

4. On chrome is there a way to make the vertical scroll bars disappear or be less intense? Could we just let the page scroll rather than scroll in pane? The sidebar should stay in place, but if the rest of the content on the page all scrolled as one I think that'd be fine.

5. Thoughts on other ways of indicating Open for the XL than something that looks like it should be clickable like a button? 
 
6. In general, thoughts on moving to some kind of different UI affordance than the same color outline for things that aren't clickable. Perhaps just a much lighter outline? This comes through on the events page, there are a lot of things that look like you should be able to click on them and there's no way to know whether you can until it tries and is unsuccessful. 

7. Sidebar for Upcoming for You
- Reservations -> Coworking Reservations (click takes you to coworking page)
- Events -> Community Events (click takes you to events page)

Can the Gear icon show on the welcome page for folks who have the right permission to edit? Takes them to the `/article/admin` link. The long term hope for admin is it's all replaced with gears and admin modules specific to the feature such that the Admin page could go away (e.g. Organizations and Academics on Admin can already be dropped from the site Admin page since they have gears). Users/Groups could be the gear icon from the personal Profile page or the About page.

ajaygandecha:
**1. On the latest news pane.**

> Can we drop the modified date (or only show it if different from created)

 ✅  The test data example is a bit poor here, but yes that is how it is currently set up! Modified should only show when it is different than created (in the test data, they are the same). But, also perfectly good with removing entirely to prevent it from being too messy. Completed in this commit!

>  Let's remove the Read button and make the card clickable? Or make elements of the card clickable (title/icon/text) and take you to details view.

I would be fine with this, although it causes problems with the interactivity of the organization chip (it is unclear that the chip is clickable). Making the title be clickable would work to solve this problem, although it would be unclear that it is clickable without an underline text decoration (which seems anti-Material)

> I think we could lose the author line from home page and leave it just to the details page

 ✅ Works with me, will keep organization though.

> When a student org is listed, hovering should show hand click icon

 ✅
 
 **2. Announcement.**
 
 > Reading an announcement details should not open a new tab
 
 ✅
 
 > The heading text feels too small relative to the heading text of a news post. Should we just use the same template as a news post but use the different highlighted color scheme?
 
Screenshot 2024-07-22 at 11 19 21 PM

Redesigned to better match the template below while also hiding unnecessary data, since there is only one announcement returned by the API and it aims to serve as a banner!

**3. Seeing some horizontal scrolling on the page at all widths**

✅ Fixed, there was some leftover padding!

**4. Scrolling**

I can see if there is a way to fix this! Currently, all of our pages use this scroll bar. In terms of having the left half free-scrolling, I originally tried this idea, however making the right sidebar sticky, grow to the correct size, scrollable, and responsive did not seem to work that well in CSS. Material's website uses the same scroll bar, but we can look towards a better site-wide solution to customize this.

**5. Open XL**

If you agree, we can use a lighter shade than `primary` so it does not appear to match the same color as the buttons used throughout the site! I would be open to `primary-container` for Open and `tertiary-container` for Closed, if you agree.

Screenshot 2024-07-22 at 11 28 44 PM

**6. Clickable Cards**

I agree - right now, both panes and outlined cards use the `surface` color scheme as recommended by material. Outlined cards use `outline (nv-50)` for outlines, and this is standard for M3.

Screenshot 2024-07-22 at 11 31 02 PM

One solution, although I am not sure how difficult this would be to pull off in Angular Material, is shown below. Material seems to recommend adding a shadow on-hover to clickable cards, basically turning them into elevated cards:

Screenshot 2024-07-22 at 11 32 58 PM

Right now, though, the site has standardized on having all elements wrapped in panes rather than directly floating on the background (for example, event cards are inside of an event pane, which is on the background). 

There is some precedent for this, although Material is not entirely consistent on how it prefers layouts:

https://m3.material.io/foundations/layout/applying-layout/large-extra-large

For our purposes, I decided it would make the most sense to stick with the pane convention, however this does raise complications on layering. If we want to try one shade different for cards placed on top of panes and standardize on that, I would be for trying it!

**7. Sidebar for Upcoming for You**

Do you mind clarifying a bit, are you mentioning adding buttons here to take users to these different pages?

**8. Gear**

✅  Yes - done!

ajaygandecha:
Re-requesting with applicable changes made!

KrisJordan (approved):
LGTM! Nice work pulling this home page together, really brings the site to the next level.

PR #543, opened 2024-07-23
Fix the Organization Creation Form

ajaygandecha:
This small PR fixes the broken organization creation form. Previously, the form checked the organization object's slug to determine if the object was new or not. Now, it correctly checks if the route includes `new` as intended.


KrisJordan (approved):
LGTM

PR #552, opened 2024-07-26
Fix Event Page when Unauthenticated

ajaygandecha:
This PR ensures that non-authenticated users are able to see the list of CSXL events. They are able to see all details, but unable to access any of the buttons (the page is view-only). For unauthenticated users, the event registrations pane is replaced with a pane that includes a nudge to sign in.


ajaygandecha:
@KrisJordan I have found some critical errors with the event creation form on the frontend when trying to create a sample featured event. After examining some things, I have decided it would be best to redesign some of the backend event service a bit - there seems to be issues with organization IDs vs slugs, and in general, I think some of these services can be approached better. To prevent conflicts, I will include it in this PR.

ajaygandecha:
@KrisJordan alright, changes pushed! This was a larger refactor, bringing events closer to some of the newer features in terms of using draft and overview models. All tests pass - I also tested this on the frontend and all features seem to be working. Also added the ability to delete events (which was not added to the frontend previously for some reason), and fixed some skewing issues with images that I saw on the main site.

PR #558, opened 2024-07-29
Add End Time to Events, Event Card Redesign, Welcome Page and Events Page Hotfix

ajaygandecha:
## Major Changes

This PR adds end times to events, allowing the organizers to specify how long the event is. 

The events card has also been redesigned to no longer use chips, instead opting for a simpler icon-and-text layout.

image

This PR solves the race condition issues on the Welcome and Event pages, as well as removes the register buttons for past events.

## Migration

@KrisJordan here are the fields to migrate:
* Rename `EventEntity.time` to `EventEntity.start`
* Add `EventEntity.end`
* Add `EventEntity.override_registration_url`


ajaygandecha:
@KrisJordan going to put the "add end time to events" PR in this branch

ajaygandecha:
@KrisJordan added!

ajaygandecha:
@KrisJordan added the ability for event organizers to add an external registration link. This hides the number of people registered, and this form of external registration is indicated on the button, like so:

Screenshot 2024-08-01 at 8 42 21 PM

The registered users table is also hidden for admins if this is enabled.

The editor has a new toggle, shown here:

Screenshot 2024-08-01 at 8 43 10 PM

When `Use the CSXL's event registration system` is not toggled, a custom registration URL field is present on the form. If on, controls to determine registration count shows.

Screenshot 2024-08-01 at 8 45 11 PM

Also, since this was not a feature before, I added a new toggle called `Add registration limit`, as shown above. If this is off, the registration limit field is hidden and the value is set to `999`. Previously, an organizer would have to type `999` for it to show up as "open to all" on the frontend. Then  If this is toggled, the registration limit form control shows up.

KrisJordan (approved):
LGTM

PR #559, opened 2024-07-29
Update course-card.widget.html

ajaygandecha:
This very small fix changes ensures that the enum string for student is updated so that the My Courses card shows correctly.


KrisJordan (approved):
LGTM

PR #567, opened 2024-08-01
Fix Course Preferences Not Displaying on Frontend Application After Edit

ajaygandecha:
This small PR ensures that the selected sections on an application form are populated when editing an application.


KrisJordan (approved):
LGTM

PR #593, opened 2024-08-19
Fix Overflow on Catalog Tables

ajaygandecha:
This small PR fixes the wide overflow on the tables in the section offerings and course catalogs tables.


KrisJordan (approved):
LGTM

PR #606, opened 2024-08-23
Add Profile Images and Location Descriptions to Office Hour Tickets

ajaygandecha:
This small PR adds profile images on the office hour ticket cards so that students can have a visual aid to find their TA, and TAs can have a visual aid to find their students.

In addition, the room location and location description now show for students whose tickets have been called.


KrisJordan (approved):
LGTM

PR #609, opened 2024-08-28
Fix External Registration Capacity Issue

ajaygandecha:
This small pull request fixes the capacity issue for external registrations, where creating an external registration set the event's capacity to 0, causing the frontend to say the event was filled.


KrisJordan (approved):
LGTM

PR #618, opened 2024-09-16
Show Room Capacity and Type on the Room Reservation Page

ellieekimm:
- Added the ability to view the capacity and room type when making a room reservation. Currently, rooms are categorized as "Pairing Rooms" (capacity = 2), "Small Group" (2 < capacity < 6), and "Large Group" (capacity > 6). 
- The formatting of the capacity and room type on the room reservation table can be changed in the future to better meet design requirements. 

**Screenshot:**

Screenshot 2024-09-16 at 1 48 21 PM

### DIFF
--- a/frontend/src/app/coworking/coworking.models.ts
+++ b/frontend/src/app/coworking/coworking.models.ts
@@ -173,7 +173,6 @@ export interface TablePropertyMap {
 }
 
 export interface ReservationMapDetails {
-  // TODO: add information to contain capacity details
   reserved_date_map: Record<string, number[]>;
   capacity_map: Record<string, number>;
   operating_hours_start: string;
--- a/frontend/src/app/coworking/coworking.models.ts
+++ b/frontend/src/app/coworking/coworking.models.ts
@@ -175,7 +175,7 @@ export interface TablePropertyMap {
 export interface ReservationMapDetails {
   reserved_date_map: Record<string, number[]>;
   capacity_map: Record<string, number>;
-  room_type_map: Record<string, string>
+  room_type_map: Record<string, string>;
   operating_hours_start: string;
   operating_hours_end: string;
   number_of_time_slots: number;
--- a/frontend/src/app/coworking/coworking.models.ts
+++ b/frontend/src/app/coworking/coworking.models.ts
@@ -175,6 +175,7 @@ export interface TablePropertyMap {
 export interface ReservationMapDetails {
   reserved_date_map: Record<string, number[]>;
   capacity_map: Record<string, number>;
+  room_type_map: Record<string, string>
   operating_hours_start: string;
   operating_hours_end: string;
   number_of_time_slots: number;
--- a/frontend/src/app/coworking/coworking.models.ts
+++ b/frontend/src/app/coworking/coworking.models.ts
@@ -173,7 +173,9 @@ export interface TablePropertyMap {
 }
 
 export interface ReservationMapDetails {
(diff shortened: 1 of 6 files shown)


frontend/src/app/coworking/widgets/room-reservation-table/room-reservation-table.widget.html
```diff
@@ -9,7 +9,10 @@
       </tr>
       @for (record of reservationsMap | keyvalue; track record.key) {
       <tr>
-        <th>{{ record.key }}</th>
+        <th>{{ record.key }}
+          <br>{{"Capacity: "}}{{ capacityMap[record.key]}}
+          <br>{{"Room Type: "}}{{ capacityMap[record.key] === 2 ? "Pairing Room" : capacityMap[record.key] < 6 ? "Small Group" : "Large Group"}}
```
ajaygandecha, line 14: I recommend removing the `Room Type` label to reduce the width of the column! On mobile devices, this would take up quite a bit of space. 

Optionally, you can also consider changing the capacity label to something like : `# seats`

kiernan-a:
Screenshot 2024-09-20 at 10 50 41 AM

Adjusted the font size/weight to decrease size of room capacity details.

frontend/src/app/coworking/widgets/room-reservation-table/room-reservation-table.widget.html
```diff
@@ -9,7 +9,10 @@
       </tr>
       @for (record of reservationsMap | keyvalue; track record.key) {
       <tr>
-        <th>{{ record.key }}</th>
+        <th>{{ record.key }}
+          <br><span class= "room-details"> {{"# seats: "}}{{ capacityMap[record.key]}}
+          <br>{{ capacityMap[record.key] === 2 ? "Pairing Room" : capacityMap[record.key] < 6 ? "Small Group" : "Large Group"}}</span>
```
ajaygandecha, line 14: Nit: I think it may be good to move this naming to the backend! My fault, I should have caught this the first time - sorry about that. Ideally, the frontend should perform as little calculations as possible! Some recommendations:

- If these names are already stored in the backend, we can just bubble this up in the `RoomEntity`.

- If not, we can either:

  - Simply move this if statement / switch case to the backend when we return this data in the map.

  - We can also add a static method in the `RoomEntity` that returns the correct name based on the capacity (this can just be done in `.toModel()`!) This functionality could also be abstracted to an enumeration or could be added to a calculated field in the entity, though that is probably overkill here

ajaygandecha (changes requested):
Looking great @ellieekimm and @kiernan-a! My apologies, there's one more nit I missed in the previous review. Once that is resolved, this can get the lgtm! Great work on this so far 🎉

frontend/src/app/coworking/coworking.models.ts
```diff
@@ -174,6 +174,8 @@ export interface TablePropertyMap {
 
 export interface ReservationMapDetails {
   reserved_date_map: Record<string, number[]>;
+  capacity_map: Record<string, number>;
+  room_type_map: Record<string, string>
```
ajaygandecha, line 178: Do we need a ; here?

ajaygandecha (approved):
LGTM - great work! 🥳

frontend/src/app/coworking/coworking.models.ts
kiernan-a: Sorry about that, just added!

PR #628, opened 2024-10-06
Change 'red' color to match M3 styling

kiernan-a:
In this PR, I switch the red shade from the default 'red' color to the M3 error40 shade (hex code: #B3261E) for the Room Reservation Table.
Screenshot 2024-10-06 at 5 20 17 PM
The above image depicts the new shade of red implemented in this PR for both the legend and the cells in the room reservation cell property map.


KrisJordan:
Can you also remove the "Page Under Construction" notice at the top of the page to this PR? Thanks!

KrisJordan (approved):
LGTM

PR #629, opened 2024-10-08
Add error message for community agreement

atoneyd:
Checks if the user has accepted the user agreement if there is an error in drafting the reservation.

Future Work: The community agreement ideally should pop up automatically on this page if the user has not accepted it.


ajaygandecha (changes requested):
Hey @atoneyd, great work on finding this issue and tackling it so far!! 🥳 

Before we merge, I think it will be better to move this validation to the backend service rather than in the component. This current solution prevents users from creating a reservation on the frontend page if they do not have a signed agreement, but does not prevent them from doing to via the docs page API or from another tool (such as a scraper).

To fix this, you can move this logic to the backend coworking service. You can create a custom exception (placed in `/backend/services/exception.py`), for example, `CoworkingCommunityAgreementException` with the same message that you have here. Then, in `main.py`, you can handle this custom exception using the middleware. Check out the [middleware error handling tutorial here[(https://github.com/unc-csxl/orientation/blob/main/backend/middleware.md)!

In the frontend, you should be able to catch this custom error message by changing the `error:` arrow function from `(_) => { ... }` to `(err) => { ... }`, then use `err` in the snacker. Feel free to repeat this process for the conflicting reservations error message as well.

This will ensure that the validation for this is end-to-end and the error message is correctly thrown to any client that plugs in! Once that is done, the PR can be merged.

backend/main.py
```diff
@@ -119,6 +119,11 @@ def permission_exception_handler(request: Request, e: UserPermissionException):
     return JSONResponse(status_code=403, content={"message": str(e)})
 
 
+@app.exception_handler(ReservationException)
+def permission_exception_handler(request: Request, e: ReservationException):
```
ajaygandecha, line 123: *Very* tiny nit (sorry!): Rename `permission_exception_handler` -> `reservation_exception_handler`

ajaygandecha (approved):
@atoneyd LGTM, great work!! 🎉 Will ping @KrisJordan to merge - thanks for getting to the bottom of this elusive bug!

PR #631, opened 2024-10-12
[Exploration] Office Hours Querying Enhancements

ajaygandecha:
This PR explores refactoring the permission check functions used in the common office hours CRUD backend service functions. The idea of these refactors is to load as little information as possible while churning over many courses, members, and sections. Some of these ideas are based off of SQLAlchemy documentation, examples, research, and some suggested optimizations from AI.

@KrisJordan and @jadekeegan, would like to hear your thoughts on this. @KrisJordan, I think to solve the problem noticed in larger courses like COMP 301, it might be helpful if you can gather any other insights on `502` errors occurring on the backend side, as well as the number of observations in the tables we are dealing with (particularly `ticket`, `office-hours` (formerly office hour events), and `section_member`).


ajaygandecha:
( @wzahrt // @atoneyd // @ItIsAndrewL), no todos or anything pressing, but pinging you all on this pull request to alert you of potential (needed) optimizations coming down the pipe to load office hours data for the TV signage, where applicable (specifically if any APIs are querying with joins between course sites, office hour events, and tickets).

ajaygandecha:
@KrisJordan, removing the `DONT MERGE` label here - if this looks good, we can look to merge.

jadekeegan (commented):
Code LGTM!

KrisJordan:
Ok, so I started looking at this and began going down a rabbit hole, but I think it's in a promising direction?

I refactored the `get_office_hours_queue` method (https://github.com/unc-csxl/csxl.unc.edu/pull/631/files#diff-e06a5209d36c36e7b58ec19ff883e5066f20bd85fdc5bdbeb9e2e3f4187fc480R52).

I really think the active queue status needs to be divorced from the sidebar stats (and this moves in that direction). Sidebar stats can be a secondary request that polls on a less frequent interval (minute?) and/or after specific actions which might impact it (like closing a ticket out). Ultimately some of these concerns might be superseded by websockets, but that likely winds up being a stream of tickets in any case.

Trying to minimize in-python traversals and logic where possible.

One thing I was curious of is why the permission detection couldn't be a little simpler? Maybe I'm underthinking things, but if you grab a membership it seems unlikely multiple memberships are at different levels (though perhaps we should assure this in some way) and once you know the level of the membership you can just do a simple set comparison rather than in the more complex SQL query.

Here we are making multiple trips to the session, but sqlalchemy and/or postgres should be caching these simple ID-based lookups to be quite fast (and there are a fixed number per query, it's not per returned row). Pulling tickets and creator(s) is done using a new to me load option (selectinload: https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#summary-of-relationship-loading-styles) -- since this is many-to-many (in theory, in practice not so much) this should be less intense than joining across, though this is probably moot in this particular case.

I want to take a pass through the other methods. I also don't want to create merge conflict hell with Jade's recurring, so I might try and get that merged in first, then rebase this again.

PR #633, opened 2024-10-12
Fix Roster Import from Canvas after Grades Added

### LINKED ISSUE #624: Cannot add roster to a course from Canvas after grades have been added
Experienced in COMP433-- attempts to upload roster failed.

The problem turned out to be an extra blank row between the Student heading and Points Possible that does not exist in exports prior to grades being added.

It leads to an [object object] box at the bottom popping up.

Suggested fix: skip over blanks in the student column.

Expected, working, export prior to grades in canvas system:
Screenshot 2024-09-24 at 1 31 05 PM

Example of problematic export with grades in system:
Screenshot 2024-09-24 at 1 30 35 PM

ajaygandecha:
This small PR adds an extra check to filter out empty rows in canvas import CSVs as recommended in #624.

*Resolves #624*

### DIFF
--- a/backend/services/academics/section_member.py
+++ b/backend/services/academics/section_member.py
@@ -186,7 +186,11 @@ class SectionMemberService:
 
         try:
             for row in reader:
-                if reader.line_num != 2 and row["Student"] != "Student, Test":
+                if (
+                    reader.line_num != 2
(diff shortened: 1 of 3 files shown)


KrisJordan (approved):
LGTM

PR #637, opened 2024-10-27
Replace login icon, show "Sign In" text on mobile over icon

### LINKED ISSUE #623: Login button looks like logout button
Especially bad on mobile - people are unable to find it

12beesinatrenchcoat:
# Summary
Replaces the "login" icon with "account_circle" on all devices, and hides the icon on mobile/thin screens (width < 768px). Also includes multiple toolbar changes to account for the width. Resolves #623.

# Details
I felt the "login" icon felt too ambiguous, so I replaced it with "account_circle". Icons like this are seen in other sites and apps as a placeholder for a profile picture when there is none (this site does that as well).
On mobile, only the icon was displayed. However, I felt that an icon by itself did not convey enough--so to completely eliminate ambiguity, the chip reads "Sign In", without the icon, when the user is signed out.
This probably shouldn't cause overflow except on significantly older or smaller devices. To account for this, I modified the toolbar:
- The left side of the toolbar is now wrapped in a div with the class "toolbar-left", which allows me to:
- Use `justify-content: space-between;` on the toolbar, also allowing me to remove the margins on `.profile-class-container`
  - The chip is now flush with the margins of the page
- In the event the screen is not wide enough, the title is clipped with ellipsis.
	- The max-width for the swap between mobile/desktop chips was increased to 768px to reduce cases of the toolbar's title being cut off.

Video (signed out):

https://github.com/user-attachments/assets/3ef25bbf-8d7f-48fb-ad93-dad8533802f3

Video (signed in):

https://github.com/user-attachments/assets/4a2154fc-8dc3-416d-8601-12446cd652c0

Additional comments: 
- Is there a formatter for this project? I would run Prettier, but `navigation.component.html` appears to have Prettier disabled, and I didn't want to mess too much with the file.
- This is my first time working with Angular; please let me know if I did something that goes against the "Angular way"! 😅

### DIFF
--- a/frontend/src/app/navigation/navigation.component.css
+++ b/frontend/src/app/navigation/navigation.component.css
@@ -94,6 +94,18 @@
   display: flex;
   flex-direction: row;
   align-items: center;
+  justify-content: space-between
+}
+
+.toolbar-left {
+  display: flex;
+  align-items: center;
+  min-width: 0;
+}
+
+#toolbar-title {
+  overflow: hidden;
+  text-overflow: ellipsis;
 }
 
 /* NOTE: This is temporary until the admin pages are refactored. */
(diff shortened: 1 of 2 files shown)


KrisJordan (approved):
LGTM thanks for contributing to the XL application!

PR #638, opened 2024-10-28
[Intermediate] Signage API

ItIsAndrewL:
## Overview

Two routes were created, `GET /api/signage/slow` and `GET /api/signage/fast`, to provide information to the frontend for the CSXL signage system in an  efficient way. These routes separate the data by update frequency—information that needs quick updates and data that can be refreshed every hour or so. Tests for these routes were added in `backend/tests/services/signage.py`.

## Slow Route

The slow route provides the five most recent news items, the next five upcoming events, the top ten users with the most check-ins (including room and CSXL check-ins) for the current month (resetting on the 1st), and the three latest announcement titles. These quantities can be adjusted in the signage service file.

## Fast Route

The fast route provides all currently active office hours using the `OfficeHoursOverview` model. This model also includes the number of people in the queue, intended for TV display.  Finally, this route also lists the currently available rooms along with seat availability in the CSXL.

## Tests

Since both routes are straightforward scripts without branching logic, we wrote a single test for each. Additional sample data was added to the `signage_data.py` file to cover scenarios not handled by existing data, such as having a leaderboard and an active room reservation.

## Additional Changes

The `to_office_hours_overview` method was moved from the service layer to the entity layer, as it was also needed in the Signage service.

### DIFF
--- a/backend/services/signage.py
+++ b/backend/services/signage.py
@@ -7,7 +7,8 @@ from sqlalchemy.orm import Session
 
 from ..database import db_session
 
-from datatime import datetime
+from datetime import datetime, timedelta
+from ..models.coworking import TimeRange
 
 from models import SignageOverviewFast, SignageOverviewSlow
 from services import ReservationService, SeatService
@@ -35,6 +36,7 @@ class SignageService:
         Gets the data for the fast API route
         """
         # Seats
+        now = datetime.now()
         walkin_window = TimeRange(
             start=now,
             end=now + timedelta(hours=6, minutes=10) # Makes sure open seats are available for 2hr walkin reservation
--- a/backend/services/signage.py
+++ b/backend/services/signage.py
@@ -1,0 +1,49 @@
+"""
+Service to collect and organize the information for CSXL Signage
+"""
+from fastapi import Depends
+from sqlalchemy import select
+from sqlalchemy.orm import Session
+
+from ..database import db_session
+
+from datatime import datetime
+
+from models import SignageOverviewFast, SignageOverviewSlow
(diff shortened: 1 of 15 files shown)


backend/services/signage.py
```diff
@@ -0,0 +1,167 @@
...
+__copyright__ = "Copyright 2024"
+__license__ = "MIT"
+
+MAX_ARTICLES = 5
+MAX_LEADERBOARD_SLOTS = 10
+MAX_EVENTS = 5
+MAX_ANNOUCEMENTS = 3
+
+
+# TODO: Create Tests for this Service
```
ajaygandecha, line 35: Nit: Can remove this todo

frontend/src/app/tv/tv.service.ts
```diff
@@ -0,0 +1,44 @@
...
+import {
+  DEFAULT_PAGINATION_PARAMS,
+  PaginationParams,
+  Paginator
+} from '../pagination';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class TvService {
```
ajaygandecha, line 28: Assume this file is not relevant to this PR anyway, but we can use `Signage` throughout in the frontend as well!

ajaygandecha (approved):
Wow, @ItIsAndrewL / @wzahrt / @atoneyd, amazing work!! Very professionally done, great work with concise code organization and logic throughout. The inclusion of the constants in the service code was a great idea. This backend code already looks mergeable to me, so.. **LGTM!** 🥳 ✨ 

*Only nit added here was for one extraneous todo comment.*

PR #640, opened 2024-11-11
Fix Queue Sort Order

### LINKED ISSUE #639: Office Hours Queue Does Not Appear in Chronological Order
The 311 team reported that the order of tickets in the queue was not chronological (so the oldest ticket didn't appear at the top). On the student side, this manifested as student positions in the queue changing from "1" to "3" to "7", causing complaints as students who had been waiting longer were not called. This issue was not found previously because the team generally hasn't had any tickets at once, but today, they had ~20 tickets all within the same time period. The timestamp is still on the tickets, but the TAs have to deliberately find the oldest one, which can be inconvenient when there are many tickets in the queue.

I suspect that when the ticket data is "re-polled", the order of each ticket record returned is random, so we may need some frontend handling to sort the tickets based on time.

ajaygandecha:
This small pull request fixes the queue sort order to ensure that it is sorted by the ticket's creation time.

*Resolves #639*

### DIFF
--- a/backend/services/office_hours/office_hours.py
+++ b/backend/services/office_hours/office_hours.py
@@ -158,11 +158,16 @@ class OfficeHoursService:
         active_ticket = active_tickets[0] if len(active_tickets) > 0 else None
 
         # Find queue position
-        queue_tickets = [
-            ticket
-            for ticket in queue_entity.tickets
-            if ticket.state == TicketState.QUEUED
-        ]
+        queue_tickets = (
+            sorted(
+                [
+                    ticket
+                    for ticket in queue_entity.tickets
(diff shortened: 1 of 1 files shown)


KrisJordan (approved):
LGTM for immediate fix; though I would wonder why not let the RBDMS handle sorting at the query stage rather than do it in Python mode. Seems like we could put some cycles into optimizing the query at some point.

jadekeegan (commented):
Aside from considering optimizations in the future, LGTM for a quick fix.

ajaygandecha:
@KrisJordan / @jadekeegan we can bubble up optimizations into #631

PR #672, opened 2025-01-17
Weather API Implementation

wzahrt:
## Digital Signage Weather Display

I have added new dividers to the header row of the signage display to show CSXL goers the current weather and upcoming announcements. Announcements currently display nothing of importance but the weather shows an icon for the type of weather outside and the current temperature. The icon displayed can be one of nine types of weather including sunny, partly cloudy, sunny and windy, partly cloudy and windy, stormy, snowy, foggy, rainy, and overcast which are pngs contained in the assets folder.

## How This Was Achieved

I used Open-Mateo as the weather API of choice as it allows the CSXL to make 10,000 free calls a day and provides useful information on the current weather conditions. I focused on the temparture_2m, weather_code, and wind_speed_10m variables as they allowed me to ascertain what kind of weather was happening based on a location (which in our case is Chapel Hill's longitude and latitude). The weather_code is based on the [WMO weather interpretation codes](https://www.nodc.noaa.gov/archive/arc0021/0002199/1.1/data/0-data/HTML/WMO-CODE/WMO4677.HTM) consisting of a table of 99 types of different weather. This weather_code variable allowed me to group many different types of rainy weather together as one icon for example. There was no weather code for windy weather so I asked made an API call for the wind_speed_10m as well and classified windy weather as having a wind speed at or over 15mph. I created functions for fetching the weather data which is called in the slow subscription and assigning the correct weather icon based on the weather_code and wind_speed_10m.

## Design Choices

I decided to use nine weather icons to visualize the appropriate types of weather that Chapel Hill receives. Although Chapel Hill does not experience fog or snow often it would be unusual not to show these weather icons if people regularly look to the signage screen to get an idea of the weather they are about to experience outside. Assigning weather icons looks like an ugly stack of if-statements at the moment. I am greatly open to suggestions to refining it.


ItIsAndrewL:
Looks like you accidently ran `npm install openmeteo` in the root directory instead of inside frontend. I'll fix this with the merge conflict fixes that I'm going to push.

frontend/src/app/signage/signage.component.ts
```diff
@@ -25,13 +59,77 @@ export class SignageComponent implements OnInit, OnDestroy {
...
+  public weatherData: any; // Store weather data
   private fastSubscription!: Subscription;
   private slowSubscription!: Subscription;
   private dateSubscription!: Subscription;
+  public current_weather_icon!: String;
 
   constructor(protected signageService: SignageService) {}
 
+  // Observable-based method to fetch weather data
+  private fetchWeatherData(): Observable<any> {
```
ItIsAndrewL, line 71: Nit: Does this belong on the service layer?

ItIsAndrewL (approved):
One nit but looks great to me! Thanks for doing this.

frontend/src/app/signage/signage.component.ts
wzahrt: Hmm that is a good question. This is the first time we are implementing an external API call into the CSXL like this. With that being said I'll make another commit such that we are receiving the weather information at the service level instead as I believe that is convention.

ajaygandecha: Agree that service level is probably the right call here

PR #685, opened 2025-01-30
Signage fix weather errors

ItIsAndrewL:
This intermediate signage PR changes the weather to use the CSXL conventions of signals instead of observables. This also reduced some complexity on the signage component level. This ended up revealing some race conditions on the `ngOnChanges` method of the slow data. The fix was found through deepseek and was simply setting the index to 0 on the change. Unfortunately this does mean that every 20 or so minutes when the slow data loads, all the slow widgets will reset off time with the spinners.

### DIFF
--- a/frontend/src/app/signage/signage.service.ts
+++ b/frontend/src/app/signage/signage.service.ts
@@ -84,7 +84,6 @@ export class SignageService implements OnInit {
   fetchWeatherData() {
     fetchWeatherApi(url, params).then((responses) => {
       const response = responses[0];
-      const utcOffsetSeconds = response.utcOffsetSeconds();
       const current = response.current()!;
 
       // Process the weather data
--- a/frontend/src/app/signage/signage.service.ts
+++ b/frontend/src/app/signage/signage.service.ts
@@ -3,6 +3,7 @@
  * from the components.
  *
  * @author Will Zahrt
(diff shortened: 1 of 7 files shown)


frontend/src/app/signage/signage.service.ts
```diff
@@ -73,32 +81,25 @@ export class SignageService implements OnInit {
...
-            temperature2m: current.variables(0)!.value(),
-            isDay: current.variables(1)!.value(),
-            weatherCode: current.variables(2)!.value(),
-            windSpeed10m: current.variables(3)!.value()
-          }
-        };
+  fetchWeatherData() {
+    fetchWeatherApi(url, params).then((responses) => {
+      const response = responses[0];
+      const utcOffsetSeconds = response.utcOffsetSeconds();
```
wzahrt, line 87: Can

wzahrt, line 87: Can remove this line of code as we don't need a timestamp for the weather anymore.

wzahrt, line 87: Can remove this line of code as we don't need a timestamp for the weather anymore.

wzahrt (approved):
Looks great Andrew! I think you can merge this immediately after you get rid of that single line in the service code.

PR #697, opened 2025-02-13
Add Digital Signage for the XL Lounge

ItIsAndrewL:
## Overview

This PR adds a signage page to the CSXL website accessible at `/signage`. The signage shows a plethora of information split up into 2 distinct categories:

1. Slow Data
2. Fast Data

### Slow Data

Slow data is the data that is only retrieved once in a while (currently it is set to refresh every 20 minutes). This data includes news articles, announcements, upcoming events, and a leaderboard. News articles and events are shown very similarly to the cards already implemented on the website. News will rotate the displayed article every 90 seconds and current events will switch the currently displayed article every 45 seconds. Changing these times is as simple as editing a number in that widget's html.

The leaderboard resets at the start of each month, encouraging ongoing engagement with CSXL. We think this would be a fun way to encourage people to venture into the CSXL and keep the leaderboard fresh and updating constantly. The leaderboard displays the top 10 people in a 5-person paginated view, rotating automatically every 30 seconds.

Announcements are only shown if one is currently published and only the first few words of the title are shown. We opted to exclude the proposed QR code here due to space constraints. We cannot make the size of announcements larger since the time at the top of the page can expand and fill the entire open space in the worst case.

### Fast Data

Fast data is the data that needs to be updated quickly (currently set to every 20 seconds) as students will need to utilize the most up to date info for the coworking seats, rooms that are currently available, and the queue sizes in office hours. Coworking seats are displayed the same way as on the website currently. We opted to display rooms in a fixed display since there are not currently any plans to add/remove any reservable rooms. Since color options were limited, we found that red provided the best contrast.

The office hours pane shows where the office hours are taking place along with the course name (+ section if the sections do not include section 001) and the queue size. This pane will rotate through columns if more than 2 are present. This was also implemented with lazy loading (bypassed if the only change was queue size) to prevent out of sync display changes.

## Testing

Backend testing was completed via pytest, with 100% coverage.

Frontend testing was completed via visual inspection and tinkering.

## Completed Screenshots

Light Mode:
light_mode

Dark Mode:
dark_mode

We appreciate your patience as we navigated unexpected challenges and tight deadlines, and we are excited to get this merged in!
- Andrew and Will


jadekeegan:
@ItIsAndrewL Merge conflicts on this branch.

ItIsAndrewL:
> @ItIsAndrewL Merge conflicts on this branch. Will review soon though!

Resolved!

frontend/src/app/signage/signage.component.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
```
ajaygandecha, line 1: Will refrain from duplicating this comment, but for now, all `.spec.ts` files for new components can be deleted.

frontend/src/app/signage/signage.component.ts
```diff
@@ -0,0 +1,125 @@
...
+  date: number = Date.now();
+  private fastSubscription!: Subscription;
+  private slowSubscription!: Subscription;
+  private dateSubscription!: Subscription;
+  private weatherSubscription!: Subscription;
+
+  constructor(protected signageService: SignageService) {}
+
+  public assginWeatherIcon(weatherData: WeatherData): string {
+    if (weatherData.weatherCode >= 95 && weatherData.weatherCode <= 99) {
```
ajaygandecha, line 50: This would be a good place to add documentation, especially for what these numbers mean! (Or add enums to codify the numbers for easier code readability and maintainability in the future)

backend/services/signage.py
```diff
@@ -0,0 +1,200 @@
...
+    ):
+        self._reservation_svc = reservation_svc
+        self._session = session
+        self._seat_svc = seat_svc
+        self.room_svc = room_svc
+
+    def office_hours_to_signage_model(
+        self, entity: OfficeHoursEntity
+    ) -> SignageOfficeHours:
+        """Converts OfficeHoursEntity into the SignageOfficeHours model"""
```
jadekeegan, line 60: **issue (blocking):** These service methods should enforce some level of permission for security. It isn't particularly sensitive data, but even just preventing unauthenticated users from accessing this information would be preferred. Possibly reach out to @KrisJordan for more insight on what level of permission we would like to require.

backend/test/services/signage_data.py
```diff
@@ -0,0 +1,118 @@
...
+    walkin=False,
+    room=None,
+    state=ReservationState.CHECKED_OUT,
+    users=[user_data.user],
+    seats=[],
+)
+
+
+def insert_fake_data(session: Session):
+    entity1 = ReservationEntity.from_model(current_reservation, session)
```
jadekeegan, line 100: **nit:** Make an array `reservations` defined as `reservations = [current_reservation, checked_out_reservation_1, ...]` (outside of this function, on line 97 or something) and convert this chunk of code to a `for` loop.

jadekeegan (changes requested):
Just looked over the backend. Looks great overall!

frontend/src/app/signage/signage.service.ts
```diff
@@ -0,0 +1,105 @@
...
+        windSpeed10m: current.variables(3)!.value()
+      });
+    });
+  }
+
+  ngOnInit(): void {
+    this.getSlowData();
+    this.getFastData();
+    this.fetchWeatherData();
+  }
```
ajaygandecha, line 104: This should be moved in the constructor, `OnInit` is typically meant for components - this could cause errors in the future.

frontend/src/app/signage/widgets/announcement-card/announcement-card.widget.ts
```diff
@@ -0,0 +1,57 @@
...
+const SECONDS_BETWEEN_CHANGE = 120;
+
+@Component({
+  selector: 'announcement-card',
+  templateUrl: 'announcement-card.widget.html',
+  styleUrls: ['announcement-card.widget.css']
+})
+export class AnnouncementCardWidget implements OnChanges, OnDestroy {
+  @Input() announcements!: string[];
+  announcement_to_display = 0;
```
ajaygandecha, line 27: Small nit: camel case preferred here : )

frontend/src/app/signage/widgets/announcement-card/announcement-card.widget.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const SECONDS_BETWEEN_CHANGE = 120;
+
+@Component({
+  selector: 'announcement-card',
+  templateUrl: 'announcement-card.widget.html',
+  styleUrls: ['announcement-card.widget.css']
+})
+export class AnnouncementCardWidget implements OnChanges, OnDestroy {
+  @Input() announcements!: string[];
```
ajaygandecha, line 26: Nit: I think it would be better to use an announcement object of sorts for announcements so that, if we want to feature more information per each announcement, we just have to update the model and API.

frontend/src/app/signage/widgets/announcement-card/announcement-card.widget.ts
```diff
@@ -0,0 +1,57 @@
...
+        });
+      } else if (
+        this.announcements.length <= 1 &&
+        this.rotating_subscription != null
+      ) {
+        this.rotating_subscription.unsubscribe();
+        this.rotating_subscription = null;
+      }
+    }
+  }
```
ajaygandecha, line 50: Opening this up to a discussion - Not required here, but I wonder if we can use a combination of a signal and an effect to handle this. Effects can run when `announcements` changes, and the active announcement can be a signal. Otherwise, I think it would be a good idea to add some documentation here so that how this works is a bit more clear - at the moment, we don't largely use `ngOnChanges`

backend/services/signage.py
ItIsAndrewL: Since we were going around the navigation for some reason I thought the Authorization header wouldn't be set, but just tested and it is. I'll work on fixing this real quick with just a basic authenticated check.

ItIsAndrewL: @jadekeegan Do you think simply adding a `subject: User = Depends(registered_user)` to the API signature is enough? I don't need the subject in the service method at all so it would just show up as an unused parameter.

ajaygandecha (changes requested):
Great work @ItIsAndrewL and @wzahrt! Very glad to see signage up and running. Jade and I are both reviewing this code, I have gone ahead and completed the initial frontend code review.

Things are looking good - only things to check and change overall in all files:
* All `.spec.ts` files can be removed for now.
* Make sure that we stick to camelCase casing for variables.
* Check out comments for where we can use signals and effects!
* **Please add docstrings / documentation** to services and a lot of the frontend code! This will make it easier for future developers. Students will see this code, and keeping to our standard of pretty good documentation (where possible) is really helpful for this. We probably shouldn't merge this until documentation is provided.

I will do a more thorough Material review once some of these comments are addressed - the only thing I am noticing right now is some inconsistencies in font sizes across pane. The content, header, and caption text sizes should remain constant across each pane. In addition, some of the headers might be a little bit large (feels large for consistent Material theming). We can look into this in a bit since it is all cosmetic. @jadekeegan may also add some other comments for design considerations!

frontend/src/app/signage/signage.component.ts
ItIsAndrewL: @wzahrt could you handle this? I also do not know what these numbers mean.

frontend/src/app/signage/widgets/announcement-card/announcement-card.widget.ts
ItIsAndrewL: Are you suggesting that we use a signal + effect to store what announcement we are displaying, or to replace the `ngOnChanges`. This `ngOnChanges` runs whenever the signal changes and is to reset the announcements widget back to a known state. This is so that if we were on announcement 2, but then it was updated to only contain 1 announcement, we don't get a null reference. I definitely think that taking a signal as the `Input()` and using an effect here would make a lot of sense instead of `ngOnChanges` but i'm not sure how we would use it to replace the observable to handle the rotating subscription.

frontend/src/app/signage/signage.model.ts
```diff
@@ -0,0 +1,98 @@
...
+export interface SignageOfficeHoursJSON {
+  id: number;
+  mode: string;
+  course: string;
+  location: string;
+  queued: number;
+}
+
+export interface SlowSignageDataJson {
+  newest_news: ArticleOverviewJson[];
```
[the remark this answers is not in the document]
ItIsAndrewL: I will change the one's that we created. The one's that we didn't create were `EventOverviewJson` and `ArticleOverviewJson`.

frontend/src/app/signage/signage.component.ts
ItIsAndrewL: @jadekeegan Since we're accessing this data from an external api I think we might run into some CORS issues if we try to access it on the backend.

backend/services/signage.py
ItIsAndrewL: After discussing with Kris, we want to leave these routes unprotected, but ensure that all the data passed doesn't include sensitive personal information. To do this, I'll create a new model for `PublicProfile` that only contains first name, last initial, and github profile picture, as this is all we need for the leaderboard and doesn't contain any sensitive personal data.

ajaygandecha: @ItIsAndrewL `PublicProfile` should already exist, feel free to use that one - and sounds good!

frontend/src/app/signage/signage.component.ts
ajaygandecha: @ItIsAndrewL makes sense - might be worth a shot to explore, otherwise (nit) it might also be good to move weather data into its own service.

frontend/src/app/signage/widgets/occupancy/occupancy.widget.ts
```diff
@@ -0,0 +1,117 @@
...
+  public title: string;
+
+  public reservable_now: boolean = false;
+  public seats_available_now: SeatAvailability[] = [];
+
+  public reservable_soon: boolean = false;
+  public next_available?: SeatAvailability;
+  public seats_available_soon: SeatAvailability[] = [];
+
+  // TODO: Handle edge case where only openings are < 1hr
```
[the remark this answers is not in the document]
wzahrt: Removed TODO and subsequent commented out code.

backend/services/signage.py
ItIsAndrewL: @ajaygandecha We are already using `PublicProfile`! Do you feel that using this without route protection exposes sensitive personal data? Currently on the XL, you cannot view someones profile without being authenticated. Kris was suggesting to make a model with even less data, i.e. only First Name, Last Initial, and github avatar.

frontend/src/app/signage/signage.component.ts
wzahrt: @ajaygandecha Weather data is already a part of the signage service, are you saying you would rather us create another service for weather data that also assigns the weather icon in the service?

frontend/src/app/signage/widgets/announcement-card/announcement-card.widget.ts
ItIsAndrewL: Read your comment again and I think you mean the latter. I'll go ahead and experiment.

backend/services/signage.py
ItIsAndrewL: Okay my solution here was to create a new `SignageProfile` model that only contains the fields we need. This way we are sending as little sensitive data to the frontend as we need. In fact, I would argue that *none* of the personal data we are sending is *sensitive*, as it is just first name, last name, and github profile picture. All of these are publicly available on the internet already: i.e. unc directory search for first & last name + github profile picture is always public.

ajaygandecha: @ItIsAndrewL sorry for the late response - the public user model is fine and is used on public facing features right now, I don't think it exposes sensitive data that is not already publicly available on the UNC Directory

ItIsAndrewL: Ah caught me too late, already implemented the stripped down version 🤪 can remove if thats better

frontend/src/app/signage/widgets/announcement-card/announcement-card.widget.ts
ItIsAndrewL: Just gave it a shot, and it seems to work. I'll go ahead and update the rest of the feature to use this instead of `ngOnChanges`

frontend/src/app/signage/signage.component.ts
ItIsAndrewL: When we move to service we could abstract out the weather icon and temperature as a signal that all components can read from.

frontend/src/app/signage/widgets/occupancy/occupancy.widget.ts
```diff
@@ -0,0 +1,117 @@
...
+      }
+    }
+  }
+
+  private initCategories(): SeatCategory[] {
+    return [
+      new SeatCategory('Sitting Desk with Monitor'),
+      new SeatCategory('Standing Desk with Monitor'),
+      new SeatCategory('Communal Area')
+    ];
```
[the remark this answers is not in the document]
ItIsAndrewL: The logic for the occupancy widget was taken from the coworking `dropin-availability-card` widget. I'll do a stripped down re-design for signage but it will need to be decoupled until we decide to update the existing `dropin-availability-card`.

frontend/src/app/signage/widgets/occupancy/occupancy.widget.html
```diff
@@ -0,0 +1,27 @@
...
+    <mat-card-title class="display-medium occupancy-title">Current Occupancy</mat-card-title>
+  </mat-card-header>
+
+  <!-- Coworking Information -->
+  <mat-card-content class="full-size-content">
+      @for (category of categories; track category) { @if
+      (category.reservable_now) {
+      <mat-list-item>
+        <span matListItemTitle class="headline-large">{{ category.title }}</span>
+        <span matListItemIcon class="display-small primary-background"><div>{{
```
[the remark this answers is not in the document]
ItIsAndrewL: I believe so, since we are using `matListItemIcon` inside of a `span`, Angular does some weird stuff with placement, and we need to be able to add a circular background, which we need another element to be able to do so. We played around with this for a while as the CSS and HTML was initially brought over from coworking's `dropin-availability-card` widget, and weren't able to get it looking similar without using this.

frontend/src/app/signage/signage.component.ts
wzahrt: New service created! Only created a new signal for weatherIcons as we can access the temperature from the weatherData signal.

frontend/src/app/signage/signage.model.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export interface FastSignageData {
+  active_office_hours: SignageOfficeHours[];
+  available_rooms: string[];
+  seat_availability: SeatAvailability[];
+}
+
+export interface WeatherData {
+  temperature2m: number;
```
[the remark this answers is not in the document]
wzahrt: Added comments for why variables are named that way.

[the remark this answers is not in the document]
wzahrt: After further consideration we will never use temperature or windspeed at higher altitudes so I removed 2m and 10m the end of those variables names.

frontend/src/app/signage/widgets/event-card/event-card.widget.html
```diff
@@ -0,0 +1,81 @@
...
+            [src]="events[shown_event].organization_icon" />
+          {{ events[shown_event].organization_name }}
+        </mat-chip-row>
+
+        @if (events[shown_event].organizers.length > 0) {
+        <mat-divider class="vertical-divider" [vertical]="true" />
+        }
+
+        <!-- Organizer Chips -->
+        @for (organizer of events[shown_event].organizers; track organizer.id) {
```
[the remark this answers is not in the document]
ItIsAndrewL: Due to sizing issues we cant use the existing widget. The css of `user-chip-list` uses exact values (`px`).

ItIsAndrewL:
image
I modeled what moving the arrows to the top of the cards would look like in figma. I'm not the biggest fan, it doesn't look very natural - and I have a feeling in my gut that we wont be able to fit the arrow next to the Leaderboard header in production.

frontend/src/app/signage/signage.model.ts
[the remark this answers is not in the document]
ajaygandecha: Unresolving since change not included yet

frontend/src/app/signage/widgets/office-hours/office-hours.widget.ts
```diff
@@ -0,0 +1,202 @@
...
+export class OfficeHoursWidget {
+  officeHours = input<SignageOfficeHours[]>([]);
+  displayOfficeHours!: SignageOfficeHours[]; // Hours that are on display currently
+  sortedHours: LocationHoursMap = {};
+  columns: Column[] = [];
+  columnsToShow: number[] = []; // Index of the columns array
+  private updater: undefined | (() => void) = undefined; // sets the new sortedHours and columns on change
+
+  constructor() {
+    effect(() => {
```
ajaygandecha, line 34: **nit**: This applies to this use of `effect` and the other uses of `effect` in other widgets - I worry `effect` may be overused in some cases where a computed signal might be better. Ultimately this will not result in major issues since this is just the signage page - but, something potentially to note. Also, it is generally best practice to have effects at the root level outside the constructor, but should not make a difference here.

frontend/src/app/signage/widgets/office-hours/office-hours.widget.ts
```diff
@@ -0,0 +1,202 @@
...
+    for (let i = 0; i < oh1.length; i++) {
+      if (
+        oh1[i].id !== oh2[i].id ||
+        oh1[i].course !== oh2[i].course ||
+        oh1[i].location !== oh2[i].location
+      ) {
+        return true;
+      }
+    }
+    return false;
```
ajaygandecha, line 145: **nit**: Can probably be reduced into one array method call

ajaygandecha (changes requested):
Looking good @ItIsAndrewL / @wzahrt! I am marking this as "Request Changes" because typography still needs fixing - I can try to do this later, but we should not merge until that gets sorted.

frontend/src/app/signage/widgets/occupancy/occupancy.widget.html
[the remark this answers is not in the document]
ajaygandecha: To add on to what @jadekeegan mentions here, I think would cause a linting error in JSX (React) and prevent building. Spans are inline elements, and `<div>` elements are block elements - only other inline elements should be placed in inline elements! Most linters would error out here, we do not have an HTML linter set up iirc - so, it would be recommended to remove. If we can replace the inner `div` with a `span` or some other type of element (or combination with CSS), that would be preferred.

This can also be part of the typography refactor needed before merge.

frontend/src/app/signage/widgets/event-card/event-card.widget.html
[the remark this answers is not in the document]
ItIsAndrewL: **Wontfix** The logic is slightly different as well, and that existing widget is using out of date constructs (i.e. `*ngFor`)

frontend/src/app/signage/widgets/occupancy/occupancy.widget.html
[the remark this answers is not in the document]
ItIsAndrewL: While this is a good point - the change would require rewriting lots of tricky css that we don't fully understand anymore, so this may just be swept under the rug.

ajaygandecha (approved):
Let's get this merged, we can fix some things as they come up later. LGTM!

jadekeegan (approved):

PR #700, opened 2025-02-16
Reorder My-Courses

embreymorton:
This PR solves the problem of inconsistencies with the order of courses on the my-courses page. This is typically an issue for TAs who have both types of courses, and the original idea was to simply have admin courses always appear first. However, the idea to mirror the GradeScope design of separating the two seemed plausible due to the amount of free space on the page.

### Changes
- Created two functions `hasInstructorCourses(term)` and `hasStudentCourses(term)` that return true or false based on if a user has at least one type of that course in the provided term.
- In the HTML, the functions are called for each type with the corresponding headers and course cards being displayed if necessary.
- Replaced duplicated "My Courses" title with the current term.
- Past courses are updated to simply list admin courses first rather than mirroring the GradeScope format.

### Testing
- Demo course data adjusted to create different scenarios of users with only instructor courses, student courses, and both types. Those changes were not pushed (I believe the OH Chime PR also adjusted the file).

#### TA with both
Screenshot 2025-02-16 at 4 15 12 PM

#### Instructor with no student courses 
Screenshot 2025-02-16 at 4 15 58 PM

#### Student with no instructor courses
Screenshot 2025-02-16 at 4 34 46 PM

### DIFF
--- a/frontend/src/app/my-courses/my-courses-page/my-courses-page.component.ts
+++ b/frontend/src/app/my-courses/my-courses-page/my-courses-page.component.ts
@@ -3,6 +3,7 @@ import { MyCoursesService } from '../my-courses.service';
 import { Router } from '@angular/router';
 import { MatDialog } from '@angular/material/dialog';
 import { CreateCourseSiteDialog } from '../dialogs/create-course-site/create-course-site.dialog';
+import { TermOverview } from '../my-courses.model';
 
 @Component({
   selector: 'app-my-courses-page',
@@ -34,4 +35,16 @@ export class MyCoursesPageComponent {
       data: this.myCoursesService.allTerms()
     });
   }
+
+  /** Returns whether or not user has a non-student role in a course during a term */
+  hasInstructorCourses(term: TermOverview): boolean {
+    return term.sites.some(course => course.role !== 'Student');
+  }
+  
+    /** Returns whether or not user has a student role in a course during a term */
+  hasStudentCourses(term: TermOverview): boolean {
+    return term.sites.some(course => course.role === 'Student');
+  }
 }
+
+
--- a/frontend/src/app/my-courses/my-courses-page/my-courses-page.component.html
+++ b/frontend/src/app/my-courses/my-courses-page/my-courses-page.component.html
@@ -23,8 +23,16 @@
         <h3 class="title-medium term-header">{{ term.name }}</h3>
         <div class="course-card-container">
           @for (course of term.sites; track course.id) {
-            <course-card [termId]="term.id" [course]="course" />
+            @if(course.role != "Student") {
+              <course-card [termId]="term.id" [course]="course" />
+            }
           }
+          @for (course of term.sites; track course.id) {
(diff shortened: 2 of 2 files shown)


ItIsAndrewL (changes requested):
Thanks for putting together such a nice PR! These changes are great and look fantastic. I know a lot of TAs will appreciate this. Personally, I'm excited just by the fact that we don't have duplicated "My Courses" anymore!!

In my testing I did find a slight bug relating to an edge case when we have multiple terms happening at together, and a student is enrolled in a course from each. You can take a look at my comments for more details.

Otherwise just a few nits. Thanks again for your work!

frontend/src/app/my-courses/my-courses-page/my-courses-page.component.ts
```diff
@@ -20,6 +20,8 @@ export class MyCoursesPageComponent {
 
   /** Whether or not to show the previous courses */
   showPreviousCourses: WritableSignal<boolean> = signal(false);
+  instructorCourses: any[] = [];
+  studentCourses: any[] = [];
```
ItIsAndrewL, line 24: **nit:** Doesn't look like you're using these fields. Go ahead and remove if so.

frontend/src/app/my-courses/my-courses-page/my-courses-page.component.ts
```diff
@@ -36,14 +38,64 @@ export class MyCoursesPageComponent {
...
-    return term.sites.some(course => course.role === 'Student');
+    /** Returns whether or not user has a student role in a course during the current terms */
+  hasStudentCourses(): boolean {
+    return this.myCoursesService.currentTerms().some((term) => {
+      return term.sites.some(course => course.role === 'Student');
+    });
+  }
+
+  /** Returns the courses where the user is an instructor during the current terms */
+  getInstructorCourses(): any[] {
```
ItIsAndrewL, line 64: **nit:** Create a new interface that extends 'CourseSiteOverview` to get rid of the `any` declaration here. You can just put this interface in this component since it probably wont be needed elsewhere.

frontend/src/app/my-courses/my-courses-page/my-courses-page.component.ts
```diff
@@ -36,14 +38,64 @@ export class MyCoursesPageComponent {
...
+    return this.myCoursesService.currentTerms().flatMap((term) => {
+      return term.sites.filter(course => course.role !== 'Student').map(course => ({
+        ...course,
+        termId: term.id
+      }));
+    });
+  }
+
+  /** Returns the courses where the user is a student during the current terms */
+  getStudentCourses(): any[] {
```
ItIsAndrewL, line 74: `any` also used here.

ItIsAndrewL (approved):
Hi Embrey! Thanks for fixing that issue! I have a few more nits for you but otherwise I think this is good to merge in. 👏 👏

embreymorton:
Sounds good, I just pushed those final changes. I appreciate your help Andrew!

Issue #596, opened 2024-08-20
Implement Digital Signage for the XL Lounge


ajaygandecha:
## Synopsis

With the new renovations, the XL hallway now has a large 65" display. These displays should be driven by special routes on the XL web application that integrate with our API and provide live, up-to-date information about XL availability, upcoming events, new articles, interesting statistics, the XL Instagram feed, etc. It is likely that the frontend for this may have to be in a slimmed down Angular application separate from the CSXL repo. We can connect to the same CSXL backend, but the frontend may just be a single page application. This helps for a few reasons:
* Less bundle size, allowing the TV to load the website properly.
* The navigation is simplified, so you can full-screen the page without worrying about the sidebar.

***Your mission, should you choose to accept it, is to implement the page (and potentially the new repo) for the live display of the XL site***.

## Acceptance Criteria

*Note: Below is an iterated version of the COMP 423 final project criteria from the Spring 2024 semester.*

- [x] Show upcoming events, including "featured" events (see the code in the events service for details on what qualifies as a featured event)

- [x] Show stats on the XL, including current coworking status (both drop-in and room reservations). Stats on office hours may also be cool!

- [x] Show latest news articles or announcements.

- [x] (Potentially ??) show a leaderboard for XL drop-ins.

- [ ] ~~Show the XL Instagram feed. For Instagram Dispaly API integration, to pull the latest post(s) see: https://developers.facebook.com/docs/instagram-basic-display-api/overview. Additionally, look toward the GitHub API integration on the user profile page for linking to GitHub. How to store keys necessary for integration with this API (whether as an environment variable or in the database) is a challenge left to the project team.~~ *Note: The development team is __NOT__ considering this as of 11/5/24*

## First Steps

If your team chooses to accept this, please list yourself as "assignees" to this ticket! Then, begin to work on a backend entity diagram design and a frontend design on Figma. Then, comment them here on this ticket for approval! From there, you can begin creating a branch for this feature.

**NOTE: This is a frontend-heavy project - the frontend design process and review will be more meticulous!**

*If you need to create a new repo, reach out to @KrisJordan!*

For approvals, we will work in this order:
1. **Backend Review:** Tests to ensure your backend is working properly.
2. **Frontend Review:** Ensures your frontend looks good!

For tips on what we will look for, check out the [Contributing Guidelines](https://github.com/unc-csxl/csxl.unc.edu/blob/main/CONTRIBUTING.md).

[...]

PR #726, opened 2025-03-16
(refactor) Digital Signage Design Refactor

ajaygandecha:
This pull request iterates on the digital signage PR from #596, refactoring components of the design to better conform to Material 3 typography standards. Other refactors, including rewriting components of CSS to make more use of flex boxes and to simplify nesting, were also completed.

## New Design

image

## Old Design

Screenshot 2025-03-15 at 11 17 17 PM

## Major Changes

* Most fonts across the page have been re-scaled based on general Material standards and precedence. The fonts across all cards are now consistent.
* The header at the top has been slightly re-arranged. The ordinal suffix has been added to the end of the date and the weather information has been right-aligned.
* The chips in the event card have been updated to be the same size as the ones in the room card.
* The leaderboard card has been redesigned to remove chips in favor of names. User images have been removed - this is to prevent numerous blank icons (over 75% of users do not have profile images). Names are also more visible.
* The dividers in the office hours card have been changed to match the color of the border of outlined cards so that they are not as jarring as the default Material color for it. Alignment of the room name has been fixed.
* CSS has been tidied up where possible - most `relative` and `absolute` positioning has been removed in favor of `flex`-based positioning.

## Testing

The design has been tested with the expected 1920px by 1080px aspect ratio for the TV - what the original design was targeting. 

## Open Questions

Through this redesign, I have been working towards the following two additional small features to be implemented, though they need further discussion:
- Showing two news articles under "Recent News" rather than one. This will help to fix the spacing issue that was prevalent before and the vertical aspect ratio on the image (which likely will scale unexpectedly for most uploaded images).
- Adding the image back to the event card above the text. There will now be enough space - I think that adding the image back will help to round that card out, I assume the main reason it was removed was to find a good default image @ItIsAndrewL?


frontend/src/app/signage/widgets/occupancy/occupancy.widget.html
```diff
@@ -1,25 +1,28 @@
 <mat-card appearance="outlined" class="full-size-pane">
   <!-- Coworking Card Header -->
   <mat-card-header class="occupancy-header">
-    <mat-card-title class="display-medium occupancy-title">Current Occupancy</mat-card-title>
+    <mat-card-title class="headline-large occupancy-title">Current Occupancy</mat-card-title>
```
wzahrt, line 4: The other mat-card-titles are "display-small" for the font size. Occupancy is the only inconsistent one.

ItIsAndrewL:
Hi @ajaygandecha thank you for these suggestions. While I think that maintaining a consistent website design across all pages makes sense for the main website, I'm not sure that our sign should be constrained by those same principles, since it will not be a part of the normal navigational flow. Instead, I think our design should be based on what we think looks *best*, and I believe that is something we've already iterated on multiple times and settled on.

That being said, a big point that you brought up was that only 75% of users on our website use profile images, and I honestly like your design better for the leaderboard (making sure to adjust to fill the space). While I would like to go straight ahead and merge that in, I'm scared we may not be able to fit large names into that same space (something I'm assuming we've tested with the chips). I'm also hopeful that once we publish this and the top users learn that they can get their GitHub image on the screen more will link their accounts!

It's for these reasons that I hope you will reconsider our original design so we can get this bad boy up on that fancy TV!

ajaygandecha:
@ItIsAndrewL thats fine, we can proceed! Great work, looking forward to seeing it on the TV

ItIsAndrewL:
I'm going to close this for now, but we can look into it later if needed.

PR #727, opened 2025-03-17
Create .env

abharti-cmd:


KrisJordan:
The .env file will contain secrets we do not want committed in the repo, so we intentionally leave it out and have establishing those secrets as part of initial set up. Thanks!

PR #776, opened 2025-06-30
chore(auth)!: allow for more flexible auth proxy use

ajaygandecha:
This PR changes the authentication logic to remove the `/auth` suffix after the origin redirect. Before this modification, since the `/auth` suffix was attached to the origin, other backends using the CSXL site as a UNC auth proxy had to implement their auth API endpoint route to have `/auth` as a suffix.

For example, the CSXL site implements the authentication route as `localhost:1560/auth` (where `origin="localhost:1560"`), while a Next.js site might have implemented the route as `localhost:3000/api/auth` (where `origin="localhost:3000"`). Both are valid, but only because the routes end with `/auth`.

There are use cases where a backend might want to connect to multiple auth providers with a route something like `localhost:8000/api/auth/unc` along with another route `localhost:8000/api/auth/google`. The current implementation of the CSXL auth logic doesn't allow for this, so this PR proposes removing the `/auth` suffix from the target. The CSXL site then can use the same authentication route `localhost:1560/auth`, only that `origin` is set to `"localhost:1560/auth"` instead of just `"localhost:1560"`. This allows for a site using multiple auth providers to set origin to `"localhost:8000/api/auth/unc"`, for example, so that the CSXL server can callback to the UNC-auth-specific endpoint.

This seems like an easier implementation than trying to modify the CSXL authentication logic to support passing along arbitrary query parameters in the provided origin (something like `localhost:8000/api/auth?provider=unc`) for which more edge-cases would arise.

This is a breaking change **(!!)** assuming other apps are currently using the CSXL as an auth proxy, which should not be the case since #774 was very recent!

Also note: this works in theory but is not testable on localhost yet because the deployment server code has to update. We can revert if something terrible goes wrong.

### DIFF
--- a/backend/api/authentication.py
+++ b/backend/api/authentication.py
@@ -40,21 +40,21 @@ In development, the proxy is not present. Instead, there are two options for aut
 1. If an unauthenticated user visits /auth in development, or staging, they are redirected
    to the production `csxl.unc.edu/auth` route with an additional query parameter `origin`.
     A. The production server authentication works as usual, but if the `origin` parameter is
-       detected alongside the SSO headers, the user will be redirected back to the `origin` 
+       detected alongside the SSO headers, the user will be redirected back to the `origin`
        server with a JWT `token` query parameter. This token is signed by the production server.
     B. Back on the development/staging server, we need to verify that the token given to the route
        was actually signed by the production server. If we did not do this, a malicious user could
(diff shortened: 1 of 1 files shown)


KrisJordan (approved):
I believe this should work and appreciate the flexibility over 3rd party URL structure. Let's give it a go.

PR #779, opened 2025-08-16
chore: add husky pre-commit commit message validator

ajaygandecha:
This PR adds a local commit validator to the CSXL repo so that commits made locally only succeed if they follow the standards of the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) format. I have noticed this as a standard in some larger open-source projects, and I have adopted this in Dabblebase and in some other projects. I think there is educational value too, since many large organizations will have a standard that they want employees to follow.

In addition, the dev container is modified to both install the necessary packages to make this work immediately after container creation, but also runs `cd frontend; npm install; cd ..` immediately as a post-creation step, ensuring node modules are up-to-date after container launch. Husky runs in a small Node environment globally in the project, but this does not interfere with the frontend node modules.

### DIFF
--- a/.commitlintrc
+++ b/.commitlintrc
@@ -1,0 +1,3 @@
+{
+  "extends": ["@commitlint/config-conventional"]
+}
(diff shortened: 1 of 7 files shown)


ajaygandecha:
Going to merge this

PR #829, opened 2025-11-24
Applicant Flagging Feature

punjarmaan:
Summary: Added flagging feature to allow HR/Coordinator to flag hiring assignments on the onboarding page

Changes:
- Add flag button on each row of hiring assignment table
- Auto-sort assignments to display flagged applicants first
- Added toggle button to filter (on FE) through all, flagged, and non-flagged applicants easily
- Add flagged boolean to hiring assignment entities to store

https://github.com/user-attachments/assets/da7e9f53-b7d4-47ff-bf99-6da9c71ed68d


ajaygandecha (commented):
Hello team, great work on this! LGTM, very nice work on the design here - I am going to request a final review from @KrisJordan to approve and ensure the migration is setup up correctly, then we can merge.

I added a quick commit for two things:
1. The optimistic updating logic for the flagging UI works well - but if there is an error, we need to roll back the change. I added quick error handling for this (and a small refactor to add a snackbar alert if an error occurs when updating, which is helpful across to the entire feature) 
2. I switched the hard-coded hex values for the flag icon to use the colors generated by material's CSS functions!

Issue #805, opened 2025-09-16
fix: background does not extend on ios 26 safari


ajaygandecha:
The XL website's background does not extend on the bottom in the new iOS 26 Safari app:

![image](https://github.com/user-attachments/assets/cd20ef70-29a2-47e8-a325-9822dcee8c25)

We need to examine whether this is the fault of the XL app or Safari itself, but other apps seem to work.

PR #844, opened 2026-04-22
fix: stabilize viewport height on iOS Safari

Thespaceblade:
Use dynamic viewport fallbacks so the app background reaches the bottom edge on iOS Safari.


Thespaceblade:
Fixes #805

